PR #1, opened 2015-03-02
changed password to a getpass prompt for all sorts of security reasons

mekhami:
Previously, your password is both visible on the command line as you type it as well as stored in your terminal history. This will prompt the user for their password using Python's built-in getpass module, which will keep it far more secure.

### DIFF
--- a/rtv/main.py
+++ b/rtv/main.py
@@ -6,6 +6,8 @@ from rtv.errors import SubmissionURLError, SubredditNameError
 from rtv.utils import curses_session
 from rtv.subreddit import SubredditPage
 from rtv.submission import SubmissionPage
+import ConfigParser
+from getpass import getpass
 
 description = """
 Reddit Terminal Viewer is a lightweight browser for www.reddit.com built into a
@@ -30,7 +32,6 @@ def main():
 
     group = parser.add_argument_group('authentication (optional)')
     group.add_argument('-u', dest='username', help='reddit username')
-    group.add_argument('-p', dest='password', help='reddit password')
 
     args = parser.parse_args()
 
@@ -38,8 +39,9 @@ def main():
         reddit = praw.Reddit(user_agent='reddit terminal viewer v0.0')
         reddit.config.decode_html_entities = True
 
-        if args.username and args.password:
-            reddit.login(args.username, args.password)
+        if args.username:
+            password = getpass()
+            reddit.login(args.username, password)
 
         with curses_session() as stdscr:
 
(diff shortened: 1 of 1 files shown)


ajclarke90:
This would be awesome. I'm not overly happy having to type my password in normal text in terminal. Thanks @mekhami

michael-lazar:
Looks good. I'd like to keep the -p flag as an option for people like me who don't care about security, and maybe change the help string to indicate that its optional and less secure. Also I think ConfigParser was renamed in python 3, that will need to be handled for compatibility. http://pythonhosted.org/six/#module-six.moves

mekhami:
I can add the -p flag back in, also I should have deleted the ConfigParser import as I was thinking about using it at first but ended up not using it.

mekhami:
Updated with those requests.

ajclarke90:
Something still doesn't seem right for me with this. If I use:

rtv -u username

It just logs me in with none of my subreddits, no password prompt pops up. Is this related to my version of python Python 2.7.6 --  64-bit.

mekhami:
Hm, that is odd. I can't replicate it. So you're essentially saying it doesn't log you in when you provide the username?

ajclarke90:
Yes, I did: 

git pull
sudo python setup.py install
rtv -u username

This then logs me in but doesn't ask for a password, and the displayed page is just the default reddit front page, similarly to if I just ran rtv without -u.

mekhami:
you shouldn't run setup.py install as sudo, but i'm not sure that would
cause this problem. i tried the exact same process and was prompted for a
password. i'll try to investigate more. can you try installing this in a
virtual environment to verify?

ajclarke90:
True, I thought the Enthought python distro defaulted to /Library but you're correct. I could have dropped the sudo. 

I can confirm the same behaviour on my python 3.3 virtualenv as well. I get rtv working but the page is the default reddit front page.

mekhami:
Hm I installed to a virtual environment and now I do get that issue. I'll do some more investigating.

ajclarke90:
Some more information if It helps. 

If I cd into the rtv directory (which should be upto date according to git status) and grep -i getpass *.py 
it returns nothing. Making it look like git isnt pulling the changes merged here.

mekhami:
Ah,you know what, it looks like Mr. Michael Lazar didn't rebase his local files with my changes before committing some other changes =P he undid my getpass changes actually. @michael-lazar can you verify this and see if you can get the changes back in?

michael-lazar:
Whoops you're right! I just pushed a fix back onto master.

PR #11, opened 2015-03-03
/top and /new

tobywhughes:
Sort of screwed up the commits because I've never actually contributed to a project before. Oh well, first time. Anyways, I added the /top and /new functionality. All I really did was check if '/' is present, and parse it if its there.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -310,14 +310,7 @@ class SubredditContent(BaseContent):
         else:
             
             if '/' in name:
-                parse = name.split('/')
-                name = parse[0]
-
-                if parse[1] == 'top':
-                    display_type = 'top'
-   
-                elif parse[1] == 'new':
-                    display_type = 'new'
+                name, display_type = name.split('/')
 
             try:
                 with loader():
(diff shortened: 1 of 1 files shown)


rtv/content.py
```diff
@@ -298,21 +298,41 @@ def __init__(self, name, submissions, loader=default_loader):
...
             return cls('Front Page', reddit.get_front_page(limit=None), loader)
 
         if name == 'all':
             sub = reddit.get_subreddit(name)
-
+        
         else:
+            
+            if '/' in name:
+                parse = name.split('/')
```
michael-lazar: How about simplifying this to 
```
name, display_type = name.split('/')
```
Then you can get rid of the temporary parse variable all together

rtv/content.py
```diff
@@ -298,21 +298,41 @@ def __init__(self, name, submissions, loader=default_loader):
...
+                    display_type = 'new'
+
             try:
                 with loader():
                     sub = reddit.get_subreddit(name, fetch=True)
             except praw.errors.ClientException:
                 raise SubredditNameError(name)
+        
+        if display_type == 'top':
+            return cls('/r/'+sub.display_name, sub.get_top_from_all(limit=None), loader)
```
michael-lazar: I think it would be good to have the type in the display name to give the user some feedback e.g.
```
'/r/'+sub.display_name+'/top'
```

michael-lazar:
Thanks! I added a couple of small comments related to code style, fix those and I'll accept the pull request.

rtv/content.py
tobywhughes: Didn't even think of that. Thanks.

tobywhughes:
There we go, should be updated.

PR #13, opened 2015-03-03
Added option to got to front page /new

tobywhughes:
Realized this was an easy thing to add. Just allows the user to do /r/new and bring up the front page's new.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -306,6 +306,9 @@ class SubredditContent(BaseContent):
 
         if name == 'all':
             sub = reddit.get_subreddit(name)
+
+        if name == 'new':
+            return cls('New', reddit.get_new(limit=None), loader)
         
         else:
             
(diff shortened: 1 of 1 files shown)


mekhami:
This functionality already exists with /r/front I believe.

rtv/content.py
```diff
@@ -306,6 +306,9 @@ def from_name(cls, reddit, name, loader=default_loader):
 
         if name == 'all':
             sub = reddit.get_subreddit(name)
+
+        if name == 'new':
```
michael-lazar: This needs to be switched around, "all" should be attached to the else block.
```
if name == 'new':
    return cls('New', reddit.get_new(limit=None), loader)

if name == 'all':
    sub = reddit.get_subreddit(name)
else:
    ...
```

michael-lazar:
Thanks for the pull request. @mekhami I think /r/front sorts the front page by 'hot', and this would sort by 'new'? Correct me if I'm wrong.

mekhami:
Yeah, sorry, I think I was reading two different threads at once and got them mixed up. Completely my misunderstanding, sorry about that.

tobywhughes:
Oh ya. My mind was thinking in else if's.

tobywhughes:
There, fixed.

PR #14, opened 2015-03-03
Preserve newlines in comments and posts

### LINKED ISSUE #12: Newlines in self-texts and comments do not render
As we can see in the screenshot above, it seems as if newlines are not rendered, causing the text to appear as a single paragraph. The post in the example above is https://www.reddit.com/r/Python/comments/2xrhq9/what_should_i_download_on_my_laptop_to_learn/.

I am using the git version as of revision 97d72a0 on Cygwin/Windows 8.1 with Python 3.2.5.

snahor:
Fix #12

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -7,6 +7,7 @@ import six
 
 from .errors import SubmissionURLError, SubredditNameError
 
+
 def clean(unicode_string):
     """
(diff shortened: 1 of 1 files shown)


michael-lazar:
Nice work!

Issue #15, opened 2015-03-03
Not possible to combine multiple subreddits


jstokes:
Generally you can have a subreddit combination /r/a+b+c and you will get stories from a, b & c.

Rtv spits out "invalid subreddit" on subreddit combinations.

(great work btw!)

tobywhughes:
I'm guessing this is an issue with PRAW. Though, looking at the docs, I still feel like this should work. While searching, I did find this

    get_multireddit(multi, *args, **kwargs)

        Return a multireddit that belongs to this user.
        Parameters:	multi – The name of the multireddit
        Returns:	Multireddit object with author=Redditor and name=multi

However, that looks like it depends on the users personal multi, not creating it with the actual address.

ddworken:
This doesn't seem to be an issue with PRAW. This can be verified by open a python console and running the following: 

```python
In [1]: import praw

In [2]: r = praw.Reddit(user_agent='my_cool_application')

In [3]: submissions = r.get_subreddit('opensource+linux').get_hot(limit=5)

In [4]: [str(x) for x in submissions]
Out[4]:
['573 :: Dell launches Ubuntu loaded machines across 500 stores in Latin America',
 '8 :: Proprietary Scumminess: The Case of Dropbox',
 '14 :: Enormous Changes to Google Summer of Code in 2015. Over 130 projects ou...',
 '18 :: Open-Source Solutions As A Business Model',
 '37 :: PuTTY 0.64 is released']
```

(Notice that you get results from both /r/opensource and /r/linux. This seems to be some problem with the from_name function in the SubredditContent class (defined in rtv/content.py).

ddworken:
After further investigation, it would appear that this might in fact be a bug with PRAW. It seems that that problem ultimately comes down to the line:

``` python
sub = reddit.get_subreddit('linux+opensource', fetch=True) 
```

This returns a 404 error, while setting fetch to False does not return a 404 error. So this can be fixed by bringing that line of code back in line with the rest of code base (which uses lazy objects). This issue is fixed by https://github.com/michael-lazar/rtv/pull/16

michael-lazar:
Thanks for investigating this, I'll check it out tonight. I put the fetch=True flag to verify that the subreddit actually existed. If you look in from_name(), there's a workaround for /r/all for the same reason. Have you tested your fix out with gibberish or invalid names? I remember having problems with those at the time.

ddworken:
[...]

PR #16, opened 2015-03-04
Allow loading of multireddits

ddworken:
Fixes https://github.com/michael-lazar/rtv/issues/15 by enabling lazy loading of subreddits. (Root cause is a PRAW bug)


michael-lazar:
See discussion on #15

Issue #17, opened 2015-03-04
Typing '/' as subreddit name breaks the application


tobywhughes:
This is the error that I get:     

     Traceback (most recent call last):
     File "main.py", line 102, in <module>
     main()
     File "main.py", line 83, in main
     page.loop()
     File "/usr/lib/python3.4/site-packages/rtv-1.0a2-py3.4.egg/rtv/subreddit.py", line 43, in loop
     self.prompt_subreddit()
     File "/usr/lib/python3.4/site-packages/rtv-1.0a2-py3.4.egg/rtv/subreddit.py", line 89, in prompt_subreddit
     self.refresh_content(name=out)
     File "/usr/lib/python3.4/site-packages/rtv-1.0a2-py3.4.egg/rtv/subreddit.py", line 67, in refresh_content
     self.reddit, name, self.loader)
     File "/usr/lib/python3.4/site-packages/rtv-1.0a2-py3.4.egg/rtv/content.py", line 332, in from_name
     sub = reddit.get_subreddit(name, fetch=True)
     File "/usr/lib/python3.4/site-packages/praw-2.1.20-py3.4.eg /praw/__init__.py", line 1017, in get_subreddit
     return objects.Subreddit(self, subreddit_name, *args, **kwargs)
     File "/usr/lib/python3.4/site-packages/praw-2.1.20-py3.4.egg/praw/objects.py", line 1346, in __init__
     subreddit_name = json_dict['url'].split('/')[2]
     TypeError: 'NoneType' object is not subscriptable
     sys:1: ResourceWarning: unclosed <socket object at 0x7f253b2b5d68>
     /usr/lib/python3.4/importlib/_bootstrap.py:2150: ImportWarning: sys.meta_path is empty

tobywhughes:
I think I figured it out. Just with the way content.py checks /. Going to submit a pull request.

michael-lazar:
Thanks. I'll leave this open since I would rather get everything squared away with a single pull request.
I have some thoughts on switching subreddits in general...

It would be nice to support for all of the following formats:
* /r/python
* /r/python/new
* /r/python/rising
* /r/python/controverisal
* /r/python/top

with optional backslash at the end
* /r/python/
* /r/python/new/
* ...

Using the same convention, we can link back to the front page. I have torn feelings about this one because /r/front already exists as a valid subreddit and using it as an alias feels kinda dirty.
* /r/front
* /r/front/new
* /r/front/rising
* ...

[...]

PR #19, opened 2015-03-04
A fix for the backslash bugs (plus front/... alias)

tobywhughes:
Once again for issue: https://github.com/michael-lazar/rtv/issues/17
Fixes everything in issue except for the multireddits.
As far as I can tell there are no more errors dealing with backslashes that won't be caught. Couldn't come up with anything anyways.

Also makes front into a sort of alias (where front/new goes to new)

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -68,6 +68,22 @@ def humanize_timestamp(utc_timestamp, verbose=False):
     years = months // 12
     return ('%d years ago' % years) if verbose else ('%dyr' % years)
 
+def validate_backslashes(name):
+    """
+    Makes sure backslashes in subreddit name work with for_name()
+    """
+
+    if name[0] == '/':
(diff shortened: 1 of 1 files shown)


rtv/content.py
```diff
@@ -309,9 +325,7 @@ def __init__(self, name, submissions, loader=default_loader):
         self._submission_data = []
 
     @classmethod
-    def from_name(cls, reddit, name, loader=default_loader):
-
-        display_type = 'normal'
+    def from_name(cls, reddit, name, loader=default_loader, display_type = 'hot'):
 
         if name == 'front':
             return cls('Front Page', reddit.get_front_page(limit=None), loader)
```
michael-lazar: You can remove this if-case now because it's being handled below. Other than that, the code looks great!

tobywhughes:
Updated

Issue #3, opened 2015-03-02
[Suggestion] Make Link Posts followable (open in browser)


mekhami:
If the post is not a self post (text post), allow the link the submission leads to to be followable, opening in the default browser.

(Sorry if you don't want suggestions here.)

michael-lazar:
Suggestions are fine! On most terminals that I've tested (at least in Ubuntu and Debian), when you hover over a link with your mouse and right click, an option appears to open the link in your browser. This is what I've been using since it's already built-in and pretty intuitive. If you want to add a keybinding to do this, I would be open to a pull request.

mekhami:
Ah yes I didn't know you could right-click it like that. That's great, I will consider adding a keybinding soon, though. Thanks :+1:

PR #20, opened 2015-03-04
Added functionality to open the submission link in the browser.

mekhami:
This will allow you to open non-self posts in a browser. It will default to opening a new browser in an existing window, or open a new window of the default browser if you do not currently have a browser open. the keystroke is 'o' which fits really nicely for key bindings and makes sense as the word 'open'.

### DIFF
--- a/rtv/subreddit.py
+++ b/rtv/subreddit.py
@@ -1,5 +1,6 @@
 import curses
 import sys
+import webbrowser
 from requests.exceptions import HTTPError
 
 from .errors import SubredditNameError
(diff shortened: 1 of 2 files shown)


mekhami:
Resolves #3

rtv/subreddit.py
```diff
@@ -1,5 +1,6 @@
 import curses
 import sys
+import webbrowser
```
michael-lazar: Nice, I didn't know this existed in the standard library

PR #22, opened 2015-03-05
to show full readme.rst in pypi

juanpabloaj:


michael-lazar:
Thanks for contributing! I tried putting this up at https://testpypi.python.org/pypi/rtv.
Looks good except for the broken image link. Do you think you can fix that?

setup.py
```diff
@@ -2,16 +2,17 @@
 #python setup.py develop --user
 #python setup.py develop --user --uninstall
 
+
+def read(*paths):
+    """ read files """
+    with open(os.path.join(*paths), 'r') as filename:
```
michael-lazar: Missing 
```python
import os
```

juanpabloaj:
sorry, now I added the

     import os

juanpabloaj:
why don't allow the image in a external service ? like imgur 

I make this with my gif in the readme.rst 

https://pypi.python.org/pypi/pip-init
https://raw.githubusercontent.com/juanpabloaj/pip-init/master/README.rst

michael-lazar:
On one hand it's nice to have the image in the project's version control so we can track the changes. On the other hand putting it on imgur makes it easier for people to share.

Here's a link http://i.imgur.com/4a3Yrov.gif
Can you update to README to point to that? I'll merge this when I get home.

juanpabloaj:
ok,  the image was updated ;)

PR #29, opened 2015-03-06
help message command added

yskmt:
Help message command added.
Hit `?` and the help message box appears.

### DIFF
--- a/rtv/utils.py
+++ b/rtv/utils.py
@@ -16,6 +16,27 @@ from .errors import EscapePressed
 
 ESCAPE = 27
 
+help_msg = """Global Commands
+  Arrow Keys or `hjkl`: Navigations.
+  `r` or `F5`         : Refresh the current page.
+  `q`                 : Quit the program.
+  `o`                 : Open the url of the selected item in the default web
+                        browser.
+  `?`                 : Show this help message.
+
+Subreddit Mode
+  Right or `Enter`    : Open the currently selected submission in a new page.
+  `/`                 : Open a prompt to switch to a different subreddit.
+
+Submission Mode
+  Right or `Enter`    : Toggle the currently selected comment between hidden
+                        and visible.
+  Left                : Exit the submission page and return to the subreddit.
+"""
+
+help_msg = help_msg.split("\n")
+
+
 class Color(object):
 
     COLORS = {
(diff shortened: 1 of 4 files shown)


rtv/utils.py
```diff
@@ -118,17 +139,29 @@ def validate(ch):
 def display_message(stdscr, message):
```
michael-lazar: You can just make the standard behavior to accept a list, instead of trying to switch between use cases. This is only being used in one other place, which would be easy enough to change to
```python
display_message(self.stdscr, ['Invalid Subreddit'])

File: rtv/utils.py [hunk 2]
```diff
@@ -118,17 +139,29 @@ def validate(ch):
...
+    if type(message) == list:
+        message_len = max(map(len, message))
+        s_row = (n_rows - len(message)) // 2
+        s_col = (n_cols - message_len - 1) // 2
+        window = stdscr.derwin(len(message) + 2, message_len + 3, s_row, s_col)
+    else:
+        message_len = len(message)
+        s_row = (n_rows - 2) // 2
+        s_col = (n_cols - message_len - 1) // 2
+        window = stdscr.derwin(10, 3, s_row, s_col)
```
michael-lazar: Where is this 10 coming from?

rtv/utils.py
```diff
@@ -16,6 +16,27 @@
...
+  Right or `Enter`    : Open the currently selected submission in a new page.
+  `/`                 : Open a prompt to switch to a different subreddit.
+
+Submission Mode
+  Right or `Enter`    : Toggle the currently selected comment between hidden
+                        and visible.
+  Left                : Exit the submission page and return to the subreddit.
+"""
+
+help_msg = help_msg.split("\n")
```
michael-lazar: I feel like this should be inside of a function. Can be as simple as
```python
HELP = """blah blah blah"""

def display_help(stdscr):
    message = HELP.split("\n")
    display_message(stdscr, message)
```
That way you don't have to import the raw string inside of subreddit.py and submission.py

rtv/utils.py
```diff
@@ -16,6 +16,27 @@
 
 ESCAPE = 27
```
michael-lazar: We need to consider combining this with the one in `main.epilog`. I actually like this version more, it's very succinct.

I'm thinking something like this in main.py
```python
from rtv.utils import HELP

epilog = """blah blah blah"""
epilog += HELP
```

michael-lazar:
Thanks for the pull request, I think this is a worthwhile feature to add. I've left you some comments.

yskmt:
I happened have some time now so I incorporated your comments. 
I agree that `epilog` and `help_msg` should be combined somehow.

michael-lazar:
I merged the help messages to a single location. I also fixed a bug where the program was crashing if the terminal was too small to display the whole help message. See 27157f400e64822b631fd40a10e96f96a14d2b1b

Issue #24, opened 2015-03-06
Invalid username / password exception not handled


michael-lazar:
Trying to start with an invalid username or password raises the following exception
```
Traceback (most recent call last):
  File "main.py", line 114, in <module>
    main()
  File "main.py", line 83, in main
    reddit.login(args.username, args.password)
  File "/usr/local/lib/python2.7/dist-packages/praw/__init__.py", line 1330, in login
    self.request_json(self.config['login'], data=data)
  File "/usr/local/lib/python2.7/dist-packages/praw/decorators.py", line 189, in wrapped
    raise error_list[0]
praw.errors.InvalidUserPass: `wrong password` on field `passwd`
```
This should be caught and handled inside of main.py, similar to how subreddit name is being done. There should be a concise error message that includes the username and starred out password e.g. `civilization_phaze_3:********`

shanhind:
I'm going to fix this tomorrow.

michael-lazar:
Fixed by shanhind

PR #30, opened 2015-03-06
Fix bug with login where invalid password throws unhandeled exception in

shanhind:
issue #24

### DIFF
--- a/rtv/main.py
+++ b/rtv/main.py
@@ -96,7 +96,7 @@ def main():
                 page.loop()
 
     except InvalidUserPass:
-        print('Invalid password for username: ' + args.username)
+        print('Invalid password for username: {}'.format(args.username)) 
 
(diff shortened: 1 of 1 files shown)


rtv/main.py
```diff
@@ -94,6 +95,9 @@ def main():
                 page = SubredditPage(stdscr, reddit, args.subreddit)
                 page.loop()
 
+    except InvalidUserPass:
+        print('Invalid password for username: ' + args.username)
```
michael-lazar: Sorry to be nitpicky, but please change the formatting style to match the rest of the errors.

See https://docs.python.org/2/library/string.html#format-examples for examples of new-style string formatting.

shanhind:
I've changed it to the format you wanted.  I understand why you'd be nitpicky about it.

PR #31, opened 2015-03-07
fix error when open subreddit in Windows 7

peterpans01:


michael-lazar:
Thanks for the pull request! Could you please give me an explanation of what this does, and why it fixes Windows in particular?

peterpans01:
From my investigation, the  function `derwin()` have the weird behavior when we just put 2 args. I try this function at two positions (first line and last line): 

    derwin_wanna_create = win_already_created.derwin(0,len("String to Test:"))

   derwin_wanna_create = win_already_created.derwin(last_line,len("String to Test:"))

1st case: it works.
2st case: curses change the positon but the string I typed didn't show up. 

So I change to give full (4) args for `derwin()`. Here this line: 

    window = self.stdscr.derwin(1, n_cols-len(prompt),n_rows-1, len(prompt))

The 2 last args are same as yours. The 2 first args just specify the derive window have exact 1 line and extend to the right corner. 

There are more information [here](https://docs.python.org/2/library/curses.html#curses.window.derwin) and [subwin()](https://docs.python.org/2/library/curses.html#curses.window.subwin)

michael-lazar:
Hmmm very interesting.  There's one other place that I'm using derwin like that

**rtv/page.py**, line 138
```python
self._content_window = self.stdscr.derwin(1, 0)
```

I'll go and change it just to be safe.

Issue #23, opened 2015-03-05
Non-Ascii characters


yskmt:
I know that encoding is a sensitive issue, but is there a reason that you convert unicode string to ascii in `clean` function? That makes non-ascii characters (Japanese, Chinese, Korean, Arabic, Emoji, etc) not rendered.

For example, adding the following line in the beginning 

    import locale
    locale.setlocale(locale.LC_ALL, '')

and changing encoding in `clean` function to something like this

    utf8_string = unicode_string.encode('utf-8', 'replace')

make non-ascii characters rendered on the console.

michael-lazar:
I think this is a fair point, I don't have anything against adding unicode. I went with ASCII because I was worried about compatibility with terminal emulators and different system configurations lacking unicode support.

michael-lazar:
https://github.com/michael-lazar/rtv/pull/33

michael-lazar:
Fixed pull request in 47ad49a189ce268d905c99229a05cfd62c7a4b0d.

However, there still appear to be some issues with python3. For example,
```
$ python2 -m rtv -l http://www.reddit.com/r/LearnJapanese/comments/2ylsz9/request_japanese_audio_textbook_audiobook_or/ --force-ascii
```
loads correctly, but

```
$ python3 -m rtv -l http://www.reddit.com/r/LearnJapanese/comments/2ylsz9/request_japanese_audio_textbook_audiobook_or/ --force-asci
```
crashes with the following error
```python
Traceback (most recent call last):
  File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/michael/Projects/rtv_project/rtv/__main__.py", line 7, in <module>
    main()
  File "/home/michael/Projects/rtv_project/rtv/main.py", line 84, in main
    page.loop()
  File "/home/michael/Projects/rtv_project/rtv/submission.py", line 28, in loop
    self.draw()
  File "/home/michael/Projects/rtv_project/rtv/page.py", line 143, in draw
    self._draw_content()
  File "/home/michael/Projects/rtv_project/rtv/page.py", line 194, in _draw_content
    attr = self.draw_item(subwindow, data, inverted)
  File "/home/michael/Projects/rtv_project/rtv/submission.py", line 95, in draw_item
    return self.draw_comment(win, data, inverted=inverted)
[...]

PR #33, opened 2015-03-08
Fixed issue #23

ragreener1:
Set utf-8 by default, if -a flag is used, then ascii is used instead.

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -1,0 +1,7 @@
+# Entry point for rtv module
+# Run by moving into the top level directory (the one with setup.py)
+# and typing "python -m rtv"
+
+from rtv.main import main
+
+main()
--- a/__main__.py
+++ b/__main__.py
@@ -1,7 +1,0 @@
-# Entry point for rtv module
-# Run by moving into the top level directory (the one with setup.py)
-# and typing "python -m rtv"
-
-from rtv.main import main
-
-main()
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -8,19 +8,30 @@ import requests
 
 from .errors import SubmissionURLError, SubredditNameError
 
+is_utf8 = True
+
 
 def clean(unicode_string):
     """
-    Convert unicode string into ascii-safe characters.
+    Checks if -a was given, if it was it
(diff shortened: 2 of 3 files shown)


__main__.py
```diff
@@ -0,0 +1,7 @@
+# Entry point for rtv module
```
michael-lazar: This is a duplicate of ``rtv/__main__.py``. Was there an intentional reason for copying this here or was it just a merge error?

rtv/content.py
```diff
@@ -8,19 +8,30 @@
...
     """
+    if not is_utf8:
+        if six.PY2:
+            ascii_string = unicode_string.encode('ascii', 'replace')
+        else:
+            ascii_string = unicode_string.encode().decode('ascii', 'replace')
 
-    if six.PY2:
-        ascii_string = unicode_string.encode('ascii', 'replace')
+            ascii_string = ascii_string.replace('\\', '')
```
michael-lazar: Why remove backslashes?

rtv/main.py
```diff
@@ -34,9 +36,11 @@ def main():
         formatter_class=argparse.RawDescriptionHelpFormatter)
     parser.add_argument('-s', dest='subreddit', help='subreddit name')
     parser.add_argument('-l', dest='link', help='full link to a submission')
+    parser.add_argument('-a', dest='force_ascii',
```
michael-lazar: What are your thoughts about changing this from ``-a`` to just ``--force_ascii``? I don't foresee this flag being used very often, and it might be good to reserve ``-a`` for something more general.

michael-lazar:
Thanks for the pull request :), I left a few comments.

__main__.py
ragreener1: this is a merge error I copied it there to run it, i'll remove that

rtv/main.py
ragreener1: Yeah that would be fine

rtv/content.py
ragreener1: this line was already here on line 22 on the original, I didn't know why it was there, so I kept it

Issue #28, opened 2015-03-06
Support for flairs


michael-lazar:
Would be nice if we could support flairs from different subreddits. Would look something like

![screenshot from 2015-03-05 20 20 33](https://cloud.githubusercontent.com/assets/5026795/6520183/5b4e240e-c375-11e4-859f-33c1e61aec75.png)

tobywhughes:
Fount this in PRAW's code overview. May be helpful:

     get_flair(subreddit, redditor) 
         Return the flair for a user on the given subreddit.

         Requires the modflair oauth scope or user/password authentication as a mod of the subreddit.

         Parameters:
        •subreddit – Can be either a Subreddit object or the name of a subreddit.
        •redditor – Can be either a Redditor object or the name of a redditor.

         Returns:
         None if the user doesn’t exist, otherwise a dictionary containing the keys flair_css_class, flair_text, and user.

shanhind:
I`m going to take a look at doing this.

shanhind:
Hey @michael-lazar, I've been looking at doing this but I'm having trouble finding where the comments and submission text are actually set / drawn to the screen.  In the submissions.py module there are functions such as draw_submission and draw_comment.  In these there are variables called text which seem to then be put into the curses screen.  I've tried adding extra text to these text variables just to see what that would affect and it actually did nothing.  Am I misunderstanding what this section of the code does?

michael-lazar:
That sounds like you're doing it correctly. Keep in mind that **subreddit.py** controls the main page, and **submission.py** controls the comments page. If you're trying to change the behavior of the main page, you need to look at **subreddit.draw_item()**. That being said, changing the text to something like
```python
text = '{author} test_flair'.format(**data)
```
should produce a visible output. If it still doesn't, my guess is that python is grabbing files from a different location than you are expecting. If so let me know and we can walk through setting everything up.

shanhind:
Yeah I think Python must be grabbing files from a different place.  Should I perhaps uninstall the system wide rtv that I installed via package manager? Or what would you suggest doing?

michael-lazar:
Here's what I've been doing lately. First make sure that your repo is up-to-date.

1. ``cd`` into the top directory (the one with ``setup.py`` and ``README.rst``)
2. type ``python -m rtv``

Explanation:
The ``-m rtv`` is shorthand for typing out ``python rtv/__main__.py``. When python sees the lines

```python
from rtv.main import main
main()
```

[...]

PR #34, opened 2015-03-11
Added functionality for flairs #28

shanhind:


rtv/subreddit.py
```diff
@@ -135,5 +135,8 @@ def draw_item(win, data, inverted=False):
         if row in valid_rows:
             text = '{author}'.format(**data)
             win.addnstr(row, 1, text, n_cols-1, curses.A_BOLD)
+            if data['flair'] != None:
```
michael-lazar: ``data['flair']`` should never be ``None``, correct?

rtv/submission.py
```diff
@@ -171,6 +171,8 @@ def draw_submission(win, data):
         attr = curses.A_BOLD | Color.GREEN
         text = '{author}'.format(**data)
         win.addnstr(row, 1, text, n_cols, attr)
+        text = '{flair}'.format(**data)
+        win.addnstr(row, 1, text, n_cols, curses.A_BOLD | Color.GREEN)
```
michael-lazar: This will overwrite the author name, as the text for both lines starts on the same row and column. If you want the text to have the same formatting as the author, I would suggest doing what you did above e.g. ``'{author} {flair}'``. If you want the flair to have different text formatting as the author, you can copy what I did below for the ``'{created} {subreddit}'`` text.

rtv/content.py
```diff
@@ -162,9 +166,14 @@ def strip_praw_submission(sub):
         data['comments'] = '{} comments'.format(sub.num_comments)
         data['score'] = '{} pts'.format(sub.score)
         data['author'] = (clean(sub.author.name) if getattr(sub, 'author')
+
                           else '[deleted]')
         data['permalink'] = clean(sub.permalink)
         data['subreddit'] = strip_subreddit_url(sub.permalink)
+        if sub.link_flair_text != None:
```
michael-lazar: You might be able to get away with a [ternary](https://docs.python.org/3.3/faq/programming.html#is-there-an-equivalent-of-c-s-ternary-operator) for this, and the one in the strip_praw_comment()
```python
data['flair'] = (clean(sub.link_flair_text) if sub.link_flair_text else "")
```
IMO it's a bit cleaner and matches the rest of the code style

rtv/subreddit.py
shanhind: Thanks for the feedback.  I'm going to go through and fix these points
tomorrow.

shanhind:
Okay I made all of the suggested changes and also changed the colour of flairs to yellow on the submission page so that they stand out from the usernames in the comments.  I was going to also make them yellow on the subreddit page but the subreddit names were already yellow so I left flairs as green on the subreddit page instead.

michael-lazar:
Thanks for the PR!

After seeing it, I'm not sure how I feel about the flairs on the subreddit page. I might play around with placement and color of the text to try to make things less cluttered. The submission page looks perfect though!

PR #38, opened 2015-03-15
add_comment function added

yskmt:
Press c to add a comment on the submission if a header is selected, or reply to a comment if the comment is selected.

Due to curses bug, textbox can only work on overwrite mode (insert mode not working).

p.s. Ignore the last pull request. I combined all the commits.

### DIFF
--- a/rtv/submission.py
+++ b/rtv/submission.py
@@ -1,9 +1,12 @@
 import curses
 import sys
+import time
+import praw.errors
 
 from .content import SubmissionContent
 from .page import BasePage
-from .utils import LoadScreen, Color, Symbol, display_help, open_browser
+from .utils import (LoadScreen, Color, Symbol, display_help, open_browser,
+                    text_input, display_message)
 
 class SubmissionPage(BasePage):
 
@@ -50,6 +53,10 @@ class SubmissionPage(BasePage):
                 self.refresh_content()
                 self.draw()
 
+            elif cmd == ord('c'):
+                self.add_comment()
+                self.draw()
+                
             elif cmd == ord('?'):
                 display_help(self.stdscr)
                 self.draw()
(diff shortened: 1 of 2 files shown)


michael-lazar:
Thanks, I'll look over this later today

rtv/submission.py
```diff
@@ -208,4 +215,48 @@ def draw_submission(win, data):
...
+            display_message(self.stdscr, ['Expand the comments first!'])
+            return
+
+        n_rows, n_cols = self.stdscr.getmaxyx()
+        box_height = n_rows/2
+
+        attr = curses.A_BOLD | Color.CYAN
+        prompt = 'Enter comment: ESC to cancel, Ctrl+g to submit'
+        prompt = '-'*((n_cols-len(prompt))/2) + prompt \
+                 + '-'*((n_cols-len(prompt)+1)/2)
```
michael-lazar: This looks really nice, good idea putting the text in the middle of the line! Have you though about using the curses ACS_HLINE instead of just a dash? I think it would look a little cleaner to have a solid line across as opposed to a dashed line.

Two ways you could do it:
1. Use [window.hline](https://docs.python.org/2/library/curses.html#curses.window.hline) and just write the text on top of the line. Downside is that ``hline`` only draws white lines.
2. Do something similar to what I did to draw the vertical bars
```python
        for y in range(n_rows):
            x = 0
            # http://bugs.python.org/issue21088
            if (sys.version_info.major,
                sys.version_info.minor,
                sys.version_info.micro) == (3, 4, 0):
                x, y = y, x
            win.addch(y, x, curses.ACS_VLINE, attr)
```

rtv/utils.py
```diff
@@ -118,7 +118,7 @@ def load_config():
 
     return defaults
 
-def text_input(window):
+def text_input(window, show_cursor=False, insert_mode=True):
```
michael-lazar: I encountered a few weird edge cases when I resized the screen while the textbox was open. 
I think a solution to this may be to cancel out of the textbox (mimic the esc key being pressed) if the window is resized. This could be easily added to the ``validate()`` function inside of ``text_input()``. What do you think?

yskmt: Yes. I think it is another `curses` bug (similar to the one described [here](http://bugs.python.org/issue13051))?

There are other issues of `curses.textpad`:
1. Infinite recursion bug when `insert_mode=True`. I actually went into the `curses` source and fixed it.
2. `stripspaces` flag. It it is true, it strips all the unnecessary spaces and line-changes, effectively destroying Reddits' markdown. If it false, every single space in textbox is returned as a part of string.

I think we have to implement a mini text editor in place.

michael-lazar:
What's the purpose of ``show_cursor=False``? It looks like on my system, neither that or ``insert_mode=True`` actually do anything. I don't get the recursion bug either. I wonder if we are pointing to  different curses libraries.

yskmt:
Here is what I have:

    Python 2.7.8 (default, Sep 29 2014, 18:17:43)
    [GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.51)] on darwin
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import curses
    >>> curses.version
    '2.2'

On my environment, `curses.curs_set(1)` is needed to show the block cursor. `curs_set(2)` does not show anything.

The flag `insert_mode=True` tries to insert a letter in place of the cursor and push back every single letter behind it by one. However, recursion is used for this process, and if the `textpad` area is too large, Python calls too many recursions and the following error occurs:

    RuntimeError: maximum recursion depth exceeded while calling a Python object

It is described [here](http://bugs.python.org/issue13051)

michael-lazar:
Here's what I have for reference:
```
Python 2.7.5 (default, Oct 13 2014, 12:50:39) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import curses
>>> curses.version
'2.2'
```
I think we should use the following settings to maintain stability across the largest number of systems:
* ``curses.curs_set(1)``. On my environment ``curs_set(1)`` and ``curs_set(2)`` look the same. I am fine sticking with ``1`` as a rule of thumb.
* ``insert_mode=False``. This would be **really** nice to have, but since it's unstable ATM I think we have to leave it out. Attempting to patch the curses module may break future python releases. The other option would be, as you said, to re-implement the textpad. However I think we would run into the same problems trying to maintain anything curses related across different os's and curses versions.
* ``strip_spaces=False``. This can be done just as well manually via ``string.rstrip(' ')``

michael-lazar:
I merged this and added some additional changes based on my comments above 9ecf7e1. If you get a chance, please try this out on the master branch and see if everything still works for you! Feel free to open up a new issue if you find anything.

Thanks again for the pull request.

yskmt:
Looks good. Sorry I didn't have time to work on it past few days.

PR #43, opened 2015-03-23
Pressing '?' crashes the client

tkajder:
I updated an import changing `docs.HELP.split('\n')` to `HELP.split('\n')` now shows help appropriately.


michael-lazar:
Thanks for the PR!

PR #45, opened 2015-03-23
Use EDITOR to write comments and replies.

noahmorrison:
Instead of pasting together your own curses textbox will all the bells and whistles that you need,
why not use what's already there?

### DIFF
--- a/rtv/docs.py
+++ b/rtv/docs.py
@@ -45,15 +45,11 @@ Submission Mode
 """
 
 COMMENT_FILE = """
---% {0} %--
-Do not edit the line above.
-Type your comment above the line.
-All text below the line will be discarded.
-
-If there is nothing but whitespace above the line
-then the comment will be discarded.
-
-Replying to {1}'s {2}
-
-{3}
+# All lines starting with a # will be ignored
+#
+# If there is only whitespace left then the comment
+# will not be posted.
+#
+# Replying to {author}'s {type}
+{content}
 """
--- a/rtv/docs.py
+++ b/rtv/docs.py
@@ -43,3 +43,17 @@ Submission Mode
   `LEFT` or `h`       : Return to subreddit mode
   `RIGHT` or `l`      : Fold the selected comment, or load additional comments
 """
+
+COMMENT_FILE = """
+--% {0} %--
+Do not edit the line above.
+Type your comment above the line.
+All text below the line will be discarded.
+
+If there is nothing but whitespace above the line
+then the comment will be discarded.
+
+Replying to {1}'s {2}
+
+{3}
+"""
--- a/rtv/submission.py
+++ b/rtv/submission.py
@@ -3,6 +3,7 @@ import sys
 import time
 import tempfile
 import os
+import uuid
 
 import praw.errors
 
@@ -11,6 +12,7 @@ from .page import BasePage
 from .helpers import clean, open_browser
 from .curses_helpers import (BULLET, UARROW, DARROW, Color, LoadScreen,
                              show_help, show_notification, text_input)
+from .docs import COMMENT_FILE
 
 __all__ = ['SubmissionPage']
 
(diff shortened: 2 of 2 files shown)


rtv/docs.py
```diff
@@ -43,3 +43,17 @@
...
+Type your comment above the line.
+All text below the line will be discarded.
+
+If there is nothing but whitespace above the line
+then the comment will be discarded.
+
+Replying to {1}'s {2}
+
+{3}
+"""
```
michael-lazar: Looks like the perfect place to use [Template String](https://docs.python.org/2/library/string.html#template-strings) formatting.

rtv/submission.py
```diff
@@ -244,33 +247,31 @@ def add_comment(self):
...
-        scol = max(0, (n_cols // 2) - (len(prompt) // 2))
-        self.stdscr.addnstr(box_height-1, scol, prompt, n_cols-scol, attr)
-        self.stdscr.refresh()
-
-        window = self.stdscr.derwin(n_rows-box_height, n_cols, box_height, 0)
-        window.attrset(Color.CYAN)
-
-        comment_text = text_input(window, allow_resize=False)
-        if comment_text is None:
+        fd, filename = tempfile.mkstemp(prefix="rtv-comment-", suffix=".txt")
```
michael-lazar: I think this can be simplified a lot using tempfile.NamedTemporaryFile

(untested)
```python
with NamedTemporaryFile(prefix="rtv-comment-", suffix=".txt") as fp:
    ...
    fp.write(comment_info)
    os.system('%s %s' % (os.getenv('EDITOR'), fp.name)
    comment_text = fp.read()
    ...
```
The file will automatically be closed at the end of the context manager and you can get rid of the try-except block.

rtv/submission.py
```diff
@@ -244,33 +247,31 @@ def add_comment(self):
...
+            with open(filename, 'w') as comment_file:
+                if data['type'] == 'Submission':
+                    info = (data['author'], 'submission', data['text'])
+                else:
+                    info = (data['author'], 'comment', data['body'])
+
+                comment_info = COMMENT_FILE.format(cid, *info)
+                comment_file.write(comment_info)
+
+            os.system('%s %s' % (os.getenv('EDITOR'), filename))
```
michael-lazar: crashes if $EDITOR is not set in the environment

rtv/docs.py
noahmorrison: They look a little clunky to me, building and passing fullblown classes, instead of a builtin string.

Also they're much slower (more than 10x slower)
```python
Python 3.4.3 (default, Feb 26 2015, 23:01:07) 
[GCC 4.9.2 20150204 (prerelease)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from string import Template
>>> from timeit import timeit
>>> t = Template('$test')
>>> def template_test():
...     t.substitute(test='hey')
... 
>>> def format_test():
...     '{0}'.format('hey')
... 
>>> timeit(stmt=template_test)
5.1952612300010514
>>> timeit(stmt=format_test)
0.4597018319982453
```

Can you give me some reasons as to why string.Template would be better?

michael-lazar: That's an interesting point. I think the timing isn't as important here because this isn't performance critical code and we're talking about a difference between 5 us and 0.5 us. My comment was more concerned about code readability. It's a lot easier for somebody skimming through to understand the context surrounding ``--% $uuid %--`` compared to ``--% {0} %--``. Or if you prefer the standard string formatting you can do ``--% {uuid} %--``, either way works for me.

BTW what's the purpose of including a uuid? It doesn't appear to be linked to anything relevant. Just brainstorming, here's what I came up with copying Git's template.

```
# Please enter a comment. Lines starting with '#' will be ignored,
# and an empty message aborts the commit.
#
# Replying to {message_type} by {author}:
#         {text}
```
Are you familiar with any other command line programs that use this technique? I looked at Mercurial and it just looks like an uglier version of Git's message.

noahmorrison: Actually git does do something similar if you use the ```-v``` flag with ```git commit```. I originally did it because I thought # was used for reddit's markdown syntax, but apparently not.

If you look at the diff that changes to using comments instead of a cutoff point, it's actually simpler to implement too.

michael-lazar: Nice

noahmorrison:
Is there anything else you want me to do to this PR?

michael-lazar:
No it looks good! I might change the wording around a little bit and set the default editor to vi or nano if it is not found. 

I've been busy, will try to get to it in the next day or two.

michael-lazar:
Merged, thank you!

I had to change the code a bit to get it to work with **gedit**. I was not able to read the changes made to the file after the editor had closed. Not sure if it was a problem with **gedit** specifically, or all non-terminal graphical editors. See https://github.com/michael-lazar/rtv/blob/master/rtv/helpers.py#L28

PR #48, opened 2015-03-30
Add version.py to MANIFEST.in

mineo:
Otherwise, version.py would not be included by `setup.py sdist`, making
the package on PyPi unusable.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include version.py


michael-lazar:
Thanks! Do you know how I can test this in the future? pip defaults to the bdist_wheel for me with both py2 and py3.

mineo:
Hallo michael-lazar:
> Thanks! Do you know how I can test this in the future? pip defaults to
> the bdist_wheel for me with both py2 and py3.

I'm not aware of anything that helps testing setup.py. To test this
patch, all I did was running `python2 setup.py sdist` which built a
tarball in dist/, which is the same one you get from PyPi. You could try
to make sure this tarball can be installed somehow (or at least its
`setup.py` can be called without an ImportError).

-- 
Wieland

PR #49, opened 2015-03-30
Replace textpad with texteditpad

yskmt:
Use `texteditpad` routine when `$EDITOR` or `$RTVEDITOR` are not defined. Need to install `texteditpad` by `pip install texteditpad`.

`texteditpad` currently supports everything `textpad` does and

1. Insert mode
2. Text wrap
3. Update on window resize
4. Scrolling

### DIFF
--- a/rtv/curses_helpers.py
+++ b/rtv/curses_helpers.py
@@ -1,9 +1,11 @@
 import os
+import sys
 import time
 import threading
 import curses
-from curses import textpad, ascii
+from curses import ascii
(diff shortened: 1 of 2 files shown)


michael-lazar:
Cool, I will try to test this out a little bit to make sure that it's stable. Is it compatible with py2 and py3?

yskmt:
Currently it only supports py2. I am working on py3 support.

yskmt:
I just implemented py3 support and uploaded to pypi.

PR #51, opened 2015-03-30
Bind enter key to open link

Brobin:
This makes navigation easier for users using the arrow keys since the `enter` key is much closer to the arrow keys than the `o` key.

### DIFF
--- a/rtv/docs.py
+++ b/rtv/docs.py
@@ -32,7 +32,7 @@ Global Commands
   `a/z`               : Upvote/downvote the selected item
   `r`                 : Refresh the current page
   `q`                 : Quit the program
-  `o`                 : Open the selected item in the default web browser
+  `ENTER` or `o`      : Open the selected item in the default web browser
   `?`                 : Show this help message
 
 Subreddit Mode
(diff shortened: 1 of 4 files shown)


Brobin:
I also refactored a bit to put the common commands in the base page. Dryer code that is easier to change in the future.

michael-lazar:
I agree, I was actually working on refactoring the command loop on a different branch.
https://github.com/michael-lazar/rtv/blob/controller/rtv/page.py#L101
Had it set up with decorators so you could do things like this:
```python
    @SubmissionController.register('o')
    def open_link(self):
        """
        Open the selected link in a webbrowser tab.
        """
        # Always open the page for the submission
        # May want to expand at some point to open comment permalinks
        url = self.content.get(-1)['permalink']
        open_browser(url)
```

I put the branch on hold at the time because I didn't want to step over a couple of other people with open pull requests. It might be worth picking back up now. The code has since gone stale, it may be easier to cherry pick changes over at this point. You're welcome to try it if you'd like.

Brobin:
Sounds good. I'll look into that

Brobin:
I'm going to close this pull request and make a new one once I've made some progress on the refactor.

Brobin:
I think I have most everything ported over. It is working on my local build. See #52

PR #52, opened 2015-03-30
Command Controllers and Command decorators

Brobin:
Ported code from the stale controller branch.

Eliminates the large while loop with nested if statements by creating a Controller class for each page that registers functions with each input command.

In addition, I added a keybinding to the `ENTER` key to open links. This make it easier for users who are navigating using the arrow keys.

### DIFF
--- a/rtv/page.py
+++ b/rtv/page.py
@@ -1,4 +1,5 @@
 import curses
+import six
 
 import praw.errors
 
@@ -126,11 +127,11 @@ class BaseController(object):
 
         func = self.character_map.get(char)
         if func is None:
-            func = Controller.character_map.get(char)
+            func = BaseController.character_map.get(char)
         if func is None:
             func = self.character_map.get(None)
         if func is None:
-            func = Controller.character_map.get(None)
+            func = BaseController.character_map.get(None)
         return func(self.instance, *args, **kwargs)
 
     @classmethod
@@ -164,9 +165,19 @@ class BasePage(object):
         self._content_window = None
         self._subwindows = None
 
+    @BaseController.register('q')
+    def exit(self):
+        sys.exit()
+
+    @BaseController.register('?')
+    def help(self):
+        show_help(self.stdscr)
+
+    @BaseController.register(curses.KEY_UP, 'k')
     def move_cursor_up(self):
         self._move_cursor(-1)
 
+    @BaseController.register(curses.KEY_DOWN, 'j')
     def move_cursor_down(self):
         self._move_cursor(1)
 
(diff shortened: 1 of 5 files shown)


rtv/page.py
```diff
@@ -115,9 +166,19 @@ def __init__(self, stdscr, reddit, content, **kwargs):
         self._content_window = None
         self._subwindows = None
 
+    @BaseController.register('q')
+    def exit(self):
+        sys.exit()
+
+    @BaseController.register('?')
+    def help(self):
+        show_help(self.stdscr)
```
michael-lazar: Missing import

Brobin: Fixed with 69a346abe8a1150f4db3eadebeb8a684885002a0

rtv/submission.py
```diff
@@ -91,19 +51,24 @@ def toggle_comment(self):
...
+    def exit_submission(self):
+        self.active = False
 
+    @SubmissionController.register(curses.KEY_F5, 'r')
+    def refresh_content(self):
         url = self.content.name
         self.content = SubmissionContent.from_url(self.reddit, url, self.loader)
         self.nav = Navigator(self.content.get, page_index=-1)
 
+    @SubmissionController.register(curses.KEY_ENTER, 10, 'o')
```
michael-lazar: What's the 10 for?

Brobin: It's the ascii value of the enter key. The curses.KEY_ENTER wasn't working. I checked the curses documentation and it says KEY_ENTER is unreliable, so I added 10 as a fallback.

michael-lazar: ah ok

rtv/page.py
michael-lazar: actually I meant show_help()!

Brobin: Oh, gotcha.

Brobin: Fixed 9a4a0b59ea19cf79ee4ae3c5a614211b00a2d3b7

rtv/page.py
```diff
@@ -115,9 +166,19 @@ def __init__(self, stdscr, reddit, content, **kwargs):
...
+
+    @BaseController.register(curses.KEY_UP, 'k')
     def move_cursor_up(self):
         self._move_cursor(-1)
```
michael-lazar: Looks like these are missing the ``self.clear_input_queue()`` that they used to have.

Basically what that does is prevent holding down the arrow keys or using the scroll wheel from building up a bunch of backlogged commands and lagging the screen once the key is released.

Brobin: On `move_cursor_up()` and `move_cursor_down()`?

michael-lazar:
Awesome, thanks. I'll let this sit on master for a few days in case anybody discovers any new bugs.

Brobin:
Sounds good. I will also continue to use it / try to break it.

PR #53, opened 2015-03-31
Fixed style to conform with PEP8

Brobin:
I noticed some of the code style was inconsistent, so I thought I'd help fix it up a little.

No functionality was changed, just the spacing and line indents, by using [autopep8](https://github.com/hhatto/autopep8)


rtv/content.py
```diff
@@ -70,9 +72,16 @@ def strip_praw_comment(comment):
...
+            data['author'] = (
+                comment.author.name if getattr(
+                    comment,
+                    'author') else '[deleted]')
+            data['is_author'] = (
+                data['author'] == getattr(
+                    comment.submission,
+                    'author'))
+            data['flair'] = (
+                comment.author_flair_text if comment.author_flair_text else '')
```
michael-lazar: These look totally mangled. If the only violation here is 80 lines, I would rather break it for the sake of readability. Or get creative with it, e.g.
```python
c_author = getattr(comment, 'author')
data['author'] = comment.author.name if c_author else '[deleted']
```

rtv/content.py
```diff
@@ -94,7 +103,10 @@ def strip_praw_submission(sub):
         data['created'] = humanize_timestamp(sub.created_utc)
         data['comments'] = '{} comments'.format(sub.num_comments)
         data['score'] = '{} pts'.format(sub.score)
-        data['author'] = (sub.author.name if getattr(sub, 'author') else '[deleted]')
+        data['author'] = (
+            sub.author.name if getattr(
+                sub,
+                'author') else '[deleted]')
```
michael-lazar: See above comment

rtv/content.py
```diff
@@ -155,9 +169,13 @@ def get(self, index, n_cols=70):
 
         elif index == -1:
             data = self._submission_data
-            data['split_title'] = textwrap.wrap(data['title'], width=n_cols-2)
-            data['split_text'] = wrap_text(data['text'], width=n_cols-2)
-            data['n_rows'] = len(data['split_title'])+len(data['split_text'])+5
+            data['split_title'] = textwrap.wrap(
+                data['title'],
+                width=n_cols -
+                2)
```
michael-lazar: Don't know why this 2 got moved to a seperate line

rtv/content.py
Brobin: Hmmm, yeah, I agree. I'll look into fixing those.

rtv/exceptions.py
```diff
@@ -1,17 +1,27 @@
 class SubmissionError(Exception):
+
```
michael-lazar: I don't think these spaces are right, see https://www.python.org/dev/peps/pep-0257/.

rtv/subreddit.py
```diff
@@ -57,9 +57,15 @@ def prompt_subreddit(self):
         attr = curses.A_BOLD | Color.CYAN
         prompt = 'Enter Subreddit: /r/'
         n_rows, n_cols = self.stdscr.getmaxyx()
-        self.stdscr.addstr(n_rows-1, 0, prompt, attr)
+        self.stdscr.addstr(n_rows - 1, 0, prompt, attr)
         self.stdscr.refresh()
-        window = self.stdscr.derwin(1, n_cols-len(prompt),n_rows-1, len(prompt))
```
michael-lazar: Don't agree with this one! I think it's a lot clearer to keep on one line 
```python
window = self.stdscr.derwin(1, n_cols - len(prompt), n_rows - 1, len(prompt))
```

rtv/content.py
```diff
@@ -155,9 +169,13 @@ def get(self, index, n_cols=70):
...
+            data['split_text'] = wrap_text(data['text'], width=n_cols - 2)
+            data['n_rows'] = len(
+                data['split_title']) + len(data['split_text']) + 5
```
michael-lazar: This one is kind of strange, I think it implies that the whole second line is wrapped by len(). I think it might be clearer to keep on a single line.

michael-lazar:
Cool. I left a few comments, feel free to speak your mind if you disagree with any of them :D.

rtv/content.py
Brobin: I like the idea of splitting them into two lines.

rtv/subreddit.py
Brobin: How about this?

```python
self.stdscr.refresh()
window = self.stdscr.derwin(1, n_cols - len(prompt),
                            n_rows - 1, len(prompt))
```

Brobin:
I went through and made some changes. Let me know what you think. Autopep8 is sometimes a little too strict. I think I found a good middle ground.

rtv/subreddit.py
michael-lazar: Works for me

Issue #39, opened 2015-03-19
Searching reddit would be nice


dominicglenn:
Searching reddit would be nice

michael-lazar:
Can you give an example use-case of how you would use the search?

Implementing this would be pretty straightforward using praw's [search](http://praw.readthedocs.org/en/latest/pages/code_overview.html#praw.__init__.UnauthenticatedReddit.search) function. Maybe instead of using ``/`` to open the subreddit prompt, you could use ``f`` (find) to open a search prompt. I don't know about any of the advanced filtering.

Brobin:
I implemented a basic search for searching a sub. You can check it out [here](https://github.com/Brobin/rtv/tree/search_subreddit)

[diff 1](https://github.com/Brobin/rtv/commit/26aab7a2cd65022e5eefdfd95abd54606de7ddc1)
[diff 2](https://github.com/Brobin/rtv/commit/f9265fd33b3bbe074890cd8fb7679cc45f2c4836)

Brobin:
Opened a pull #56

michael-lazar:
Yea make a pull request please. A couple of things I noticed:

- Will need a special case for the front page
- Might want to make the prompt a different color to differentiate from the subreddit prompt. green or yellow maybe?
- Should be catching specific exceptions in your **try-except** block
- Are multireddits searchable? E.g. ``/r/python+programming``

PR #56, opened 2015-03-31
Subreddit Search implemented

Brobin:
I implemented searching of a subreddit using praw's search function. Simply type `f` to open the search prompt. If no results were found, there is a popup message.

Also moved prompt_input to it's own method in the BasePage so it can be used for future prompts.

#39

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,6 +70,7 @@ In subreddit mode you can browse through the top submissions on either the front
 
 :``►`` or ``l``: View comments for the selected submission
 :``/``: Open a prompt to switch subreddits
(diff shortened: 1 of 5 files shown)


Brobin:
From issue #39 

> Will need a special case for the front page

Yep, just noticed this

> Might want to make the prompt a different color to differentiate from the subreddit prompt. green or yellow maybe?

Does it need to be different? I thought since you'd only have one at a time and they'd each have descriptions that the same color would be fine.

> Should be catching specific exceptions in your try-except block

Will do

> Are multireddits searchable? E.g. /r/python+programming

Yes they are.

PR #58, opened 2015-04-01
Added posting a Submission (text post)

Brobin:
Reused a bit of the code from commenting.

The only issue is the page refresh. I think the submit method is taking longer so it's not refreshing the content correctly. Maybe we could make another loader to run until the submission is done?

Also, probably save this for next version, 1.2 perhaps.

### DIFF
--- a/rtv/subreddit.py
+++ b/rtv/subreddit.py
@@ -1,6 +1,7 @@
 import curses
 import time
 import requests
+import praw
 
 from .exceptions import SubredditError
 from .page import BasePage, Navigator, BaseController
@@ -102,6 +103,10 @@ class SubredditPage(BasePage):
             show_notification(self.stdscr, ['Can\'t post to a multireddit'])
             return
 
+        name = name.strip(' /')  # Strip leading and trailing backslashes
+        if name.startswith('r/'):
+            name = name[2:]
+
         # Open the submission window
         submission_info = SUBMISSION_FILE.format(name=name)
         curses.endwin()
@@ -114,7 +119,7 @@ class SubredditPage(BasePage):
             return
         try:
             title, content = submission_text.split('\n', 1)
-            show_notification(self.stdscr, [title])
+            self.reddit.submit(name, title, text=content)
         except praw.errors.APIException as e:
             show_notification(self.stdscr, [e.message])
         else:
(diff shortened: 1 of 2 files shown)


rtv/subreddit.py
```diff
@@ -89,6 +91,41 @@ def open_link(self):
...
 
+    @SubredditController.register('p')
+    def post_submission(self):
+        # Abort if user isn't logged in
+        if not self.reddit.is_logged_in():
+            show_notification(self.stdscr, ["Login to reply"])
+            return
+
+        name = self.content.name
+        if '+' in name:
```
michael-lazar: I think that there are other cases where this will fail as well, eg ``/r/all``. You could do 
```python
sub = self.reddit.get_subreddit(name)
```
to check that the subreddit is valid. It's an extra API call but I think it would be worth it. Also you wouldn't have to guess about the name because you could pass in the subreddit object.
```python
self.reddit.submit(sub, title, text=content)
```

Brobin: What would `get_subreddit` return for `/r/all`?

Brobin: Or would something like this be better?

```
elif 'all' in name:
    show_notification(self.stdscr, ['Can\'t post to /r/all'])
    return

Comment (michael-lazar) [inline rtv/subreddit.py]: Yes it will fail.

```python
try:
    sub = self.reddit.get_subreddit(name)
except APIException:
    # Handle case where subreddit doesn't exist
```
Go with your way if you think it's cleaner though. Also don't forget about the modifiers like ``/r/front/new`` and ``/r/all/controverisal``

Brobin:
Ah, yep. Modifiers are still an issue (/new, /controversial, etc.)

Brobin:
I fixed the subreddit name for the modifieres (/new, etc.).

It seems that `get_subreddit` for All and Front return `/r/all` and `/r/front`, so that is not an issue.

rtv/subreddit.py
```diff
@@ -89,6 +91,42 @@ def open_link(self):
...
+        curses.endwin()
+        submission_text = open_editor(submission_info)
+        curses.doupdate()
+
+        # Abort if there is no content
+        if not submission_text:
+            curses.flash()
+            return
+        try:
+            title, content = submission_text.split('\n', 1)
```
michael-lazar: Will raise ``ValueError: need more than 1 value to unpack`` if the text is only one line

Brobin:
I just noticed we missed adding the documentation for this. Do you wanna add it real quick to master? It doesn't seem worth opening a whole pull request for.

Brobin:
Attached new docs to #74

PR #59, opened 2015-04-01
comment/reply edit and delete functions added

yskmt:
There are some common parts in `add_comment`, `edit_comment` and `delete_comment`, which can be refactored. Also, sometimes the comment deletion and edition are not immediately updated. Is it a PRAW problem?

### DIFF
--- a/rtv/submission.py
+++ b/rtv/submission.py
@@ -9,7 +9,7 @@ from .page import BasePage, Navigator, BaseController
 from .helpers import clean, open_browser, open_editor
 from .curses_helpers import (BULLET, UARROW, DARROW, Color, LoadScreen,
                              show_notification, text_input)
-from .docs import COMMENT_FILE
+from .docs import COMMENT_FILE, COMMENT_EDIT_FILE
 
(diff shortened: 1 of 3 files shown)


michael-lazar:
My guess would be that Reddit is caching the submission on the server side and only updates the cache every so often. 

I agree there is some work that could be done to refactor this. There is a lot of code that will be shared between the SubmissionPage and the SubredditPage.

michael-lazar:
For deleting, how about a confirmation prompt
```
Are you sure you want to delete this post? (y/n):
```

rtv/submission.py
```diff
@@ -112,9 +119,74 @@ def add_comment(self):
...
+        curses.endwin()
+        comment_text = open_editor(comment_info)
+        curses.doupdate()
+
+        if not comment_text:
+            curses.flash()
+            return
+        try:
+            if data['type'] == 'Submission':
+                data['object'].edit(comment_text)
```
Brobin: Redundant?

```python
data['object'].edit(comment_text)
```

yskmt:
In order to rebase to the newest master, I will close this pull request.

PR #61, opened 2015-04-02
Login with `u` key

yskmt:
I don't think we have to change any data structures after logging in.

### DIFF
--- a/rtv/page.py
+++ b/rtv/page.py
@@ -226,17 +226,43 @@ class BasePage(object):
         except praw.errors.LoginOrScopeRequired:
             show_notification(self.stdscr, ['Login to vote'])
 
-    def prompt_input(self, prompt):
+    @BaseController.register('+')
+    def login(self):
+        username = self.prompt_input('Enter username: ')
+        password = self.prompt_input('Enter password: ', hide=True)
+
+        if (username == '') or (password == ''):
+            curses.flash()
+            return
+        try:
+            self.reddit.login(username, password)
+            
+        except praw.errors.InvalidUserPass:
+            show_notification(self.stdscr,
+                              ['Invalid password for username: {}'.format(username)])
+        else:
+            show_notification(self.stdscr,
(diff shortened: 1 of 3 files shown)


michael-lazar:
Cool! I don't like the ``+`` key though.. how about ``u``?

yskmt:
yeah `u` is fine. I just realized that PRAW log you out when you fail to log in. I will come up with a fix.

rtv/page.py
```diff
@@ -226,17 +226,43 @@ def downvote(self):
...
+        username = self.prompt_input('Enter username: ')
+        password = self.prompt_input('Enter password: ', hide=True)
+
+        if (username == '') or (password == ''):
+            curses.flash()
+            return
+        try:
+            self.reddit.login(username, password)
+            
+        except praw.errors.InvalidUserPass:
```
Brobin: Is this thrown only when the password is wrong, or if either the username doesn't exist or password is wrong?

yskmt: @Brobin Yeah I think that checks username/password combination. If the username does not exists, then it always fails.

yskmt:
The only way I can think of to not get logged out is to first check username/password with dummy log in. Then upon successful dummy log in, RTV tries to log in with the real internal variable.

Also, I changed the log in behavior so that if there is no username specified, then RTV logs out.

michael-lazar:
How about this:

- If you are logged in, pressing ``u`` will prompt ``Log out? (y/n):`` and then log out
- If you are not logged in, pressing ``u`` will prompt for username/password

So to switch accounts, you need to press ``u`` twice, once to log out and once to log back in. That way the user is protected against accidentally logging out.

Brobin:
This seems like a good way to handle this. :+1:

michael-lazar:
I cleaned the code up a little bit and merged this. Thanks!

I think that the hidden password prompt is still a little weird right now. Re-sizing the terminal while entering the password breaks the login. It would be great if you could find a cleaner solution for this.

PR #63, opened 2015-04-02
added program title to terminal title

Brobin:
Displays the name and version in the terminal title. Nothing too fancy, just helps when you have 5 terminals open :D

Tested on linux. The first method works for linux and OSX. The second should work for Windows..

### DIFF
--- a/rtv/__main__.py
+++ b/rtv/__main__.py
@@ -15,6 +15,7 @@ from .curses_helpers import curses_session
 from .submission import SubmissionPage
 from .subreddit import SubredditPage
 from .docs import *
+from .__version__ import __version__
 
 __all__ = []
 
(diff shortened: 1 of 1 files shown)


rtv/__main__.py
```diff
@@ -68,6 +69,13 @@ def main():
     args = command_line()
     local_config = load_config()
 
+    # set the terminal title
+    title = 'rtv {0}'.format(__version__)
+    try:
+        sys.stdout.write("\x1b]2;{0}\x07".format(title))
```
michael-lazar: Do you have a source that explains how this works?

Brobin: I found it on the [arch wiki](https://bbs.archlinux.org/viewtopic.php?id=85567).

Windows method (and linux again) on this [stackoverflow post](http://stackoverflow.com/questions/2330393/how-to-set-the-program-title-in-python#answer-2330596)

michael-lazar: Interesting. I think based on that link the windows method should be 
```python
os.system('title {}'.format(title))
```

Brobin: My mistake I'll fix that real quick.

Brobin:
I also just make it check `os.name` for the windows version because I don't think `sys.stdout` would throw an exception in Windows.

PR #66, opened 2015-04-02
Manipulate comment

yskmt:
Refactoring and rebase of https://github.com/michael-lazar/rtv/pull/59

Comment add/edit/delete are merged in one function `manipulate_comment`.

### DIFF
--- a/rtv/submission.py
+++ b/rtv/submission.py
@@ -71,14 +71,11 @@ class SubmissionPage(BasePage):
         url = self.content.get(-1)['permalink']
         open_browser(url)
 
-    def manipulate_comment(self, command):
-        """
-        Handles add, delete, and edit comment/reply.
-        """
-        
+    def get_comment_data(self):
+
         if not self.reddit.is_logged_in():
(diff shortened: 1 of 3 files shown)


rtv/submission.py
```diff
@@ -71,15 +71,14 @@ def open_link(self):
         url = self.content.get(-1)['permalink']
         open_browser(url)
 
-    @SubmissionController.register('c')
-    def add_comment(self):
+    def manipulate_comment(self, command):
```
michael-lazar: I don't think that this currently works well as a single function. There are too many 
```python
if command == 'c':
    ...
elif command == 'e':
   ...
else
```
style blocks. You should factor out some of the "special cases" and place them back into their parent functions. I'll try to give some examples.

michael-lazar: Actually the more I look at this, the more I think that it should just be 3 seperate functions.

yskmt:
either way is fine for me. If I split them into three functions. Then it looks like my newest commit.

michael-lazar:
Merged in #83

PR #68, opened 2015-04-03
toggle hidden comment with arrow left or h

Brobin:
I can't count how many times I've pressed the left arrow to un-hide a comment and have been kicked back out the the sub. I figured if I'm having this problem others might as well.

This simple change checks to see if the comment is hidden on `KEY_LEFT` or `h` key presses and if it is, it un-hide it instead of exiting to the subreddit

EDIT: You won't hurt my feelings if this doesn't get added, I just thought it might be useful :)

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -229,6 +229,10 @@ class SubmissionContent(BaseContent):
         else:
             raise ValueError('% type not recognized' % data['type'])
 
+    def is_hidden_comment(self, index):
+        data = self.get(index)
+        return data['type'] == 'HiddenComment'
+
 
 class SubredditContent(BaseContent):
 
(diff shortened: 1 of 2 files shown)


michael-lazar:
I think core problem here is the RIGHT isn't a very intuitive key for toggling comments. Every other directional key is used to navigate between comments and pages. The reason I did it this way was because it made it easy for me to browse everything with only one hand on the arrow keys. 

For a while I was also using SPACE to toggle comments. It actually works really well if you're using **hjkl**, but is a little awkward combined with the arrow keys. I would say try that out for a while and let me know how it feels.

Brobin:
Sounds good. It agree is still kind awkward to hide comments (that move left) by pressing the right arrow.

Brobin:
Maybe the `CTRL` key would be a good mapping. Combined with `SPACE` that would be convenient for both arrow key users and `hjkl` users.

Brobin:
closing for now

PR #73, opened 2015-04-03
Read config from XDG_CONFIG_HOME (Fix #65)

### LINKED ISSUE #65: XDG Specification
Follow [XDG specs](http://standards.freedesktop.org/basedir-spec/basedir-spec-0.7.html#introduction) and read the config file in the following order:

* `$XDG_CONFIG_HOME/rtv/rtv.conf`
* `~/.config/rtv/rtv.conf` (in case XDG_CONFIG_HOME is not set)
* `.rtv` (for compatibility reasons)

Ram-Z:
Config file is read from the first of these two locations:
* $XDG_CONFIG_HOME/rtv/rtv.cfg
* ~/.rtv

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,13 +91,11 @@ In submission mode you can view the self text for a submission and browse commen
 Configuration
 -------------
 
-RTV will read a configuration file located at **~/.rtv**.
+RTV will read a configuration file located at ``$XDG_CONFIG_HOME/rtv/rtv.cfg`` or ``~/.config/rtv/rtv.cfg`` if ``$XDG_CONFIG_HOME`` is not set.
 This can be used to avoid having to re-enter login credentials every time the program is launched.
 Each line in the file will replace the corresponding default argument in the launch script.
 
 Example config:
-
-**~/.rtv**
 
 .. code-block:: ini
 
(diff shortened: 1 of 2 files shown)


Ram-Z:
I've removed the mention of `~/.rtv` in the README, as usually you only want a single place for the config file. I'd be happy to add it back in if you prefer to support both location.

README.rst
```diff
@@ -91,14 +91,12 @@ In submission mode you can view the self text for a submission and browse commen
 Configuration
 -------------
 
-RTV will read a configuration file located at **~/.rtv**.
+RTV will read a configuration file located at ``$XDG_CONFIG_HOME/rtv/rtv.cfg``.
```
michael-lazar: Change to ``~/.config/rtv/rtv.cfg``? I'm worried that a lot of people won't know what $XDG_CONFIG_HOME stands for, or that it defaults to ``~/.config`` if not set.

Ram-Z: Good point... added mention of default location if `XDG_CONFIG_HOME` is not set.

michael-lazar:
Looks good, thanks!

PR #83, opened 2015-04-06
Edit/Delete Submissions and Comments

Brobin:
A lot of this logic was basically the same for submissions and comments, so I decided to put it in the BasePage. 

I cherry picked some code from #66 Thanks to @yskmt

### DIFF
--- a/rtv/page.py
+++ b/rtv/page.py
@@ -257,22 +257,24 @@ class BasePage(object):
         Delete a submission or comment.
         """
         data = self.content.get(self.nav.absolute_index)
-        if data['author'] == self.reddit.user.name:
-            prompt = 'Are you sure you want to delete this? (y/n):'
-            char = self.prompt_input(prompt)
-            if char == 'y':
-                try:
-                    data['object'].delete()
-                    show_notification(self.stdscr, ['Deleted'])
-                except praw.errors.APIException as e:
-                    show_notification(self.stdscr, [e.message])
-                else:
-                    time.sleep(0.5)
-                    self.refresh_content()
-            else:
-                show_notification(self.stdscr, ['Delete cancelled'])
-        else:
+        if data['author'] != self.reddit.user.name:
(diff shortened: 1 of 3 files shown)


rtv/page.py
```diff
@@ -251,6 +252,70 @@ def login(self):
         else:
             show_notification(self.stdscr, ['Welcome {}'.format(username)])
 
+    @BaseController.register('d')
+    def delete(self):
+        """
+        Delete a submission or comment.
+        """
+        data = self.content.get(self.nav.absolute_index)
+        if data['author'] != self.reddit.user.name:
```
michael-lazar: data['author'] may not be a valid field

rtv/page.py
```diff
@@ -251,6 +252,70 @@ def login(self):
...
+            show_notification(self.stdscr, ['You can\'t edit this'])
+            return
+
+        if data['type'] == 'Submission':
+            subreddit = self.reddit.get_subreddit(self.content.name)
+            sub = str(subreddit).split('/')[2]
+            sub_file = '{content}' + SUBMISSION_FILE
```
michael-lazar: I think it would be better to have this be the default ``SUBMISSION_FILE`` template, and set ``content=''`` in the other case.

Brobin: sounds good

rtv/page.py
Brobin: I was finding a `KeyError` when I ran it

Brobin:
Overall question, do you think it would be best to just do nothing if a user tries to edit or delete something that is not theirs or not editable? I feel like we have a lot of extra error notifications.

rtv/page.py
Brobin: fixed

michael-lazar:
Merged, thanks @Brobin and @yskmt for the contributions! I fixed a few edge cases, for example you shouldn't be able to delete the submission that you're currently viewing. I also factored out the common error handling between the post/edit/delete functions.

Issue #27, opened 2015-03-06
Bindings for pageup and pagedown


michael-lazar:
I would like to add support for scrolling up and down a full page at a time.

tobywhughes:
What about going from parent comment to parent comment? I feel like a full page is sort of arbitrary in terms of reddit comments. However, it would be more useful to be able to jump from the highest level parent comment to the next when going through posts like comments on askreddit where a single parent post can have hundreds and hundreds of replies.

michael-lazar:
Normally when I'm browsing Reddit I will fold the comment threads once I'm done with them. That's at least one way to jump to the next parent comment. I assume that's what reddit expects you to do, at least I haven't found anything to tell me otherwise. I'm going to start working on this. I'll try out both ways and see what feels more natural.

michael-lazar:
Implemented by the awesome @yskmt. Key bindings are `n` for pageup and `m` for pagedown. I'll update the README before the next release goes out.

PR #86, opened 2015-04-10
page up/down implemented

yskmt:
To address issue https://github.com/michael-lazar/rtv/issues/27

`m` and `n` keys to page up and down. 

It is implemented for both submission and subreddit mode, but I think it is more useful to browse submissions in subreddit mode.

### DIFF
--- a/rtv/page.py
+++ b/rtv/page.py
@@ -84,6 +84,35 @@ class Navigator(object):
 
         return valid, redraw
 
+    def move_page(self, direction, n_windows):
+        """Move the page and cursor down (positive direction) or up (negative
+        direction)"""
+
+        # top of submission page: act as normal move
+        if self.absolute_index < 0:
+            valid, redraw = self.move(direction, n_windows)
+        else:
+            # first page
+            if self.absolute_index < n_windows and direction < 0:
+                self.page_index = -1
+                self.cursor_index = 0
+                self.inverted = False
+
+                # not submission mode: starting index is 0
+                if not self._is_valid(self.absolute_index):
+                    self.page_index = 0
+                
+                valid = True
+            else:
+                self.page_index += n_windows*direction
+                valid = self._is_valid(self.absolute_index)
+                if not valid:
+                    self.page_index -= n_windows*direction
+
+            redraw = True
+
+        return valid, redraw
+            
     def flip(self, n_windows):
         "Flip the orientation of the page"
 
@@ -384,29 +413,8 @@ class BasePage(object):
         self._remove_cursor()
 
         if page_ud:
-
-            # top of submission page:  act as normal move
-            if self.nav.absolute_index < 0:
-                valid, redraw = self.nav.move(direction, len(self._subwindows))
-            # top of submission comment: go back to the title
-            elif self.nav.absolute_index == 0 and direction < 0:
-                valid, redraw = self.nav.move(direction, len(self._subwindows))
-            else:
-                # first page: goes up to the top
-                if self.nav.absolute_index < len(self._subwindows)\
-                   and direction < 0:
-                    self.nav.page_index = 0
-                    self.nav.cursor_index = 0
-                    self.nav.inverted = False
-                    valid = True
-                else:
-                    self.nav.page_index += len(self._subwindows)*direction
-                    valid = self.nav._is_valid(self.nav.absolute_index)
-                    if not valid:
-                        self.nav.page_index -= len(self._subwindows)*direction
-
-                redraw = True
-                
+            valid, redraw = self.nav.move_page(direction,
+                                               len(self._subwindows))
         else:
             valid, redraw = self.nav.move(direction, len(self._subwindows))
 
--- a/rtv/page.py
+++ b/rtv/page.py
@@ -188,6 +188,16 @@ class BasePage(object):
         self._move_cursor(1)
         self.clear_input_queue()
 
+    @BaseController.register('m')
+    def move_cursor_page_down(self):
+        self._move_cursor(1, page_ud=True)
+        self.clear_input_queue()
+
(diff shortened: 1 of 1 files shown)


rtv/page.py
```diff
@@ -372,8 +382,11 @@ def _remove_cursor(self):
     def _move_cursor(self, direction):
 
         self._remove_cursor()
-
-        valid, redraw = self.nav.move(direction, len(self._subwindows))
+        if abs(direction) > 1:
+            direction = (len(self._subwindows)-1)*direction/abs(direction)
```
michael-lazar: Doing it this way will only work if you expect `len(_subwindows)` to always be the same. For example, say your cursor is at the bottom of the screen and you page down. If the next submission in the list is very large and takes up most of the screen, you will scroll right past it. 

What we really need to do is make it so that paging down takes the next submission on the list (the one after the  bottom submission on the screen) and makes it the new top submission on the screen. And likewise for paging up. Does that make sense? I think this will be a bit more complicated to pull off.

yskmt:
@michael-lazar I re-implemented the page up/down functions from the scratch.

I think I implemented the desired behavior. Page down moves the next-to-the-bottom submission or reply to the top while keeping the cursor position. Page up does the opposite. There are some edge conditions that makes things complicated, but it should be fine.

michael-lazar:
I was playing around with this and got an error after moving the cursor to the middle of the page and trying to page down

```python
  File "/home/michael/Projects/rtv_project/rtv/__main__.py", line 117, in main
    page.loop()
  File "rtv/subreddit.py", line 45, in loop
    self.controller.trigger(cmd)
  File "rtv/page.py", line 175, in trigger
    return func(self.instance, *args, **kwargs)
  File "rtv/page.py", line 240, in move_cursor_page_up
    self._move_cursor(-1, page_ud=True)
  File "rtv/page.py", line 518, in _move_cursor
    self._add_cursor()
  File "rtv/page.py", line 497, in _add_cursor
    self._edit_cursor(curses.A_REVERSE)
  File "rtv/page.py", line 526, in _edit_cursor
    window, attr = self._subwindows[self.nav.cursor_index]
IndexError: list index out of range
```

yskmt:
I think the problem was keeping the cursor position after the page up/down. For example, if the cursor position is 5 from the top, then it stays at the position 5 after page down. However, if that position does not exists after page up/down, it causes error. I fixed that problem.

The other possibility is just move the cursor to the top-most position after page up/down.

michael-lazar:
I am ok with moving the cursor to the top of the page after page up/down, especially if it simplifies the code.

yskmt:
Actually, moving the cursor to the top after page up/down does not simplify the code because I still have to check the edge condition. So it is a matter of preference. I personally like keeping the cursor position as it is.

rtv/page.py
```diff
@@ -388,6 +440,10 @@ def _edit_cursor(self, attribute=None):
         if self.nav.absolute_index < 0:
             return
 
+        # Don't allow the cursor to go over the subwindow number
```
michael-lazar: This check should be integrated into `move_page()`, not here. We should never allow an invalid cursor position.

rtv/page.py
```diff
@@ -369,11 +416,16 @@ def _add_cursor(self):
...
 
-    def _move_cursor(self, direction):
+    def _move_cursor(self, direction, page_ud=False):
 
         self._remove_cursor()
 
-        valid, redraw = self.nav.move(direction, len(self._subwindows))
+        if page_ud:
+            valid, redraw = self.nav.move_page(direction,
+                                               len(self._subwindows))
```
michael-lazar: I think using `len(self._subwindows)` is the wrong thing to do here. Here's an example:

Say the screen looks like this with '>' being the cursor.
```
  post 9
  post 10
  post 11
  post 12
  post 13
> post 14
  post 15
```

When you page up, the screen will move up 6 posts.

```
  post 3
  post 4
  post 5
  post 6
  post 7
> post 8
  post 9
```

Now consider what happens if `post 5` is realllly large and takes up 4 times the amount of room on the screen. After paging up, your screen will look like this
```
  post 3
  post 4
  post 5




> post 6
```
This is a big problem because we accidentally skipped right over `post 7` and `post 8`! Paging down will have a similar problem if `nav.inverted == True` because in that case, the page is relative to the bottom of the screen. Does that make sense?

michael-lazar:
Hey I left some more comments. Sorry for taking so long on this. I think it's important that we get this code 100% correct, otherwise there are going to be a bunch of hidden bugs and edge cases. I appreciate the effort that you're putting in!

rtv/page.py
yskmt: One way to avoid it is always redraw from the bottom (nav.inverted==True) when page up and always redraw from the top (nav.inverted==False) when page down. In this way I don't have to think about the length of the contents of each post.

yskmt: As for a separate issue, what should we do if some post doesn't fit in the window and truncated? In the current implementation, as long as a part of some post is within the window, page up/down skip that post.

michael-lazar: I agree with your first point. Check out `navigator.flip()`, you might be able to leverage that to change the screen orientation before paging. 

For the second issue, how about moving `max(1, len(subwindows) - 1)` instead of `len(subwindows)`. That way any partial post will still be visible. In the rare case where there isn't a partial post, you end up with bottom post moved to the top of the screen. I see this as a positive because it will help users reorient their position on the screen.

michael-lazar:
Is this currently working? I hadn't noticed that you updated it.

yskmt:
Yes. Check out the latest commit.

michael-lazar:
Merged! Looks great, thanks for cleaning up the code :+1:

PR #95, opened 2015-04-27
History Log across multiple sessions

Brobin:
How this works

1. Load the contents of the history file into a set
2. As a link is visited, add it to the set
3. On exit (crtl+c or q), write a max of 200 elements of the set to the file

### DIFF
--- a/rtv/history.py
+++ b/rtv/history.py
@@ -1,0 +1,37 @@
+import os
+
+
+__all__ = ['load_history', 'save_history']
+
+
+def history_path():
+    """
+    Create the path to the history log
+    """
+    HOME = os.path.expanduser('~')
+    XDG_CONFIG_HOME = os.getenv('XDG_CACHE_HOME', os.path.join(HOME, '.config'))
+    path = os.path.join(XDG_CONFIG_HOME, 'rtv', 'history.log')
+    return path
+
+
+def load_history():
+    """
+    Load the history file into memory if it exists
+    """
+    path = history_path()
+    if os.path.exists(path):
+        with open(path) as history_file:
+            return set([line.replace('\n', '') for line in history_file])
+    return set()
+
+
+def save_history(history):
+    """
+    Save the visited links to the history log
+    """
+    path = history_path()
+    with open(path, 'w+') as history_file:
+        for i in range(200):
+            if not history:
+                break
+            history_file.write(history.pop() + '\n')
--- a/rtv/history.py
+++ b/rtv/history.py
@@ -23,7 +23,8 @@ def load_history():
     path = history_path()
     if os.path.exists(path):
         with open(path) as history_file:
-            history = [line.strip() for line in history_file]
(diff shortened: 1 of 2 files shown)


rtv/history.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def load_history():
+    """
+    Load the history file into memory if it exists
+    """
+    path = history_path()
+    if os.path.exists(path):
+        with open(path) as history_file:
+            return set([line.replace('\n', '') for line in history_file])
```
michael-lazar: You can use a set comprehension here. It's a small change but I think it looks a little cleaner and saves an intermediate step.

```python
return {line.strip() for line in history_file}
```

rtv/history.py
```diff
@@ -0,0 +1,37 @@
...
+    return set()
+
+
+def save_history(history):
+    """
+    Save the visited links to the history log
+    """
+    path = history_path()
+    with open(path, 'w+') as history_file:
```
michael-lazar: I think this will fail if the directory (e.g. `~/.config/rtv/`) doesn't exist. check out `os.makedirs`.

rtv/history.py
```diff
@@ -0,0 +1,37 @@
...
+        for i in range(200):
+            if not history:
+                break
+            history_file.write(history.pop() + '\n')
```
michael-lazar: Using `pop()` on a set returns an arbitrary element, and does not guarantee order. I'm pretty sure you're going to have to use a list for that. One way to solve this would be to keep two data structures internally, one to store the order and one to test membership. Here's something that comes to mind. Not saying you have to do this, it's just the first idea I had.

```python
class History(object):

    def __init__(self, ...):
        self._hlist = []
        self._hset = {}

    def __contains__(self, item):
        return (item in self._hset)

    def add(self, item):
        self._hlist.append(item)
        self._hset.add(item)

    def load(...)
    def save(...)
    def path(...)
```

Another nice side-effect of switching to a class is that you could take your 
```
@atexit.register
save()
``` 
call and stick it right inside of the `load()` method.

Brobin: The ordering issue was something I was thinking about. I'll definitely look into this. This seems like a good solution.

Brobin:
I refactored it to use an `OrderedSet` class with an internal list and set to maintain order but check for membership in O(1).

I elected to leave loading and saving as separate functions in order to treat `OrderedSet` as a data structure,

Brobin:
I just did some more manual testing, and it looks like the history is not preserved when switching subreddits.

michael-lazar:
Yea I don't see any reason why that would matter. Is the list getting cleared all together, or is it getting reset to whatever is in the history file?

Brobin:
I can't seem to replicate it. Maybe I messed with the file in my editor.

michael-lazar:
Code looks great, nice implementation!

PR #97, opened 2015-04-29
Colors config

arnoutpierre:
Support colors themes:

+ Read default colors in `rtv/config.py` (black & white)
+ Load theme located in `$XDG_CONFIG_HOME/rtv/rtv-colors.cfg`
+ Use defined colors in page / subreddit / submission

Check `rtv-colors.dist.cfg` for example

Screenshot:
RTV Colors


rtv/config.py
```diff
@@ -1,5 +1,26 @@
 """
 Global configuration settings
 """
-
 unicode = False
+
+default_colors = {
+    'CommentsText':      ( -1, 0 ),
+    'Header':            ( -1, 0 ),
```
Brobin: Shouldn't this be `HeaderBg` and `HeaderFg`?

```
  File "/py3venv/lib/python3.4/site-packages/rtv-1.3-py3.4.egg/rtv/__main__.py", line 135, in <module>
  File "/py3venv/lib/python3.4/site-packages/rtv-1.3-py3.4.egg/rtv/__main__.py", line 118, in main
  File "/py3venv/lib/python3.4/site-packages/rtv-1.3-py3.4.egg/rtv/subreddit.py", line 43, in loop
  File "/py3venv/lib/python3.4/site-packages/rtv-1.3-py3.4.egg/rtv/page.py", line 460, in draw
  File "/py3venv/lib/python3.4/site-packages/rtv-1.3-py3.4.egg/rtv/page.py", line 469, in _draw_header
AttributeError: type object 'Color' has no attribute 'HeaderBg'
```

Brobin:
This shouldn't mess with existing installations. When I try running it without a config file, I'm getting lots of AttributeErrors. the defualt color keys in config.py don't match the config keys.

Brobin:
Additionally, I think if there isn't an config file, it should use the terminal's default colors. I have my terminal colors set, but by overriding everything with black and white, you get something like this.

arnoutpierre:
I just fix missing keys in `config.py`, my bad.
I will try to fill `default_colors` hash with current used colors today.

For the terminal set colors, it's not the best way because you can not dissociate your global terminal theme an the rtv theme, that's why I have forked this repo.

Brobin:
I think if you revert the `default_colors` dict to use the `curses` colors like before, then it will be fine.

arnoutpierre:
I'm thinking to turn this PR to a real **themes** support, and not only colors.
My idea is to:

+ Switch colors config path to `$XDG_CONFIG_HOME/rtv/colors/`
+ Allow to use one or several colors files by reading all files in this folder and use `sections`
+ Add an entry `theme` in `rtv.cfg` to defined theme to use

Examples:
`$XDG_CONFIG_HOME/rtv/colors/defaults.cfg`:
```
[dark]
HeaderBg=x,y
... stuff ...

[light]
HeaderBg=x,y
... stuff ...
```

`$XDG_CONFIG_HOME/rtv/colors/monokai.cfg`:
```
[monokai]
... stuff ...
```

`$XDG_CONFIG_HOME/rtv/colors/solarized`:
```
[solarized]
... stuff ...
```

`$XDG_CONFIG_HOME/rtv/rtv.cfg`
```
... stuff ...
theme=monokai
```

Maybe it's the good time to add this feature, it's not a big deal but it's pretty useful when you want to test themes, share them ect ..

Tell me what you think about that please

shaggytwodope:
@arnoutpierre the ability to make the background color "default" or transparent of the title would be wonderful. It's annoying for all transparent/dark themed kinda guys like me lol. But I love the direction of adding full on themes.

Maybe something to the effect of `$XDG_CONFIG_HOME/rtv/colors/solarized.theme` ? and option to include/use said theme with `[themename]` in the config.

arnoutpierre:
@shaggytwodope currently you should be able to turn any part in transparent.

Currently `Title` is used as key for thread titles, so if this is what you want to turn into transparent juste do: `Title=<font-color>,-1`

If you speak about the `Header` background (the bar where is displayed current subreddit and username) do: 
```
HeaderBg=<font-color>,-1
HeaderFg=<font-color>,-1
UserName=<font-color>,-1
```

I just realize that is not really smart, I have to check how do this better.
Another cool thing will be to allow theme to manage global background color.

I will try to do all these things tonight (french hour) !

arnoutpierre:
Hum, last thing:

What about adding bold / italic / underline and these stuff in theme config ?
Technical is not a big deal too ..

arnoutpierre:
I just added theme support like explained before and update defaults colors as required

rtv/curses_helpers.py
```diff
@@ -4,10 +4,12 @@
 import curses
 from curses import textpad, ascii
 from contextlib import contextmanager
+from six.moves import configparser
 
 from .docs import HELP
 from .helpers import strip_textpad
 from .exceptions import EscapeInterrupt
+import config
```
Brobin: from . import config

rtv/__main__.py
```diff
@@ -47,6 +46,8 @@ def load_config():
     if 'unicode' in defaults:
         defaults['unicode'] = config.getboolean('rtv', 'unicode')
 
+    rtvconfig.theme = defaults['theme']
```
Brobin: We should make sure `"theme"` is in `defaults`, otherwise we'll get a `KeyError`.

```python
if 'theme' in defaults:
    rtvconfig.theme = defaults['theme']
else:
    rtvconfig.theme = None
```

Brobin:
The colors look better! I like where this is going, @arnoutpierre.

arnoutpierre:
Well, I think I should not add more code.

The styles (bold / underline) are now supported in themes. Both (like colors) are defined by integer (`BOLD`: `2097152`, `UNDERLINE`: `131072`).

To be simpler to use two aliases are available (`BOLD`, `UNDERLINE`). Whatever you use aliases or not, it's required to define style at the end of list.

`config.default_colors` do not support aliases.

Here is my personal theme (called monokai-extended who's not a good name): https://gist.github.com/arnoutpierre/0226e087b9e6988d1a0f

rtv/curses_helpers.py
```diff
@@ -183,9 +185,77 @@ def init(cls):
...
+                        values.append( int( value ) )
+                    except:
+                        pass
+
+            colors[ key ] = values
+        return colors
+
+    @classmethod
+    def load_themes(cls):
+        """ Load all themes in config paths, return theme indexed hash of colors  """
```
Brobin: Is it really necessary to load all of the themes? Wouldn't it make more sense to simply look for the one defined in the config, and if it doesn't exist, fall back to the defaults?

rtv/curses_helpers.py
```diff
@@ -183,9 +185,77 @@ def init(cls):
...
+        themes = cls.load_themes()
+        # get theme to use if exists
+        theme  = themes[theme] if theme in themes.keys() else {}
+
+        # compute colors hash
+        colors = {}
+        # loop in theme entries
+        for key in theme:
+            values = []
+            for value in theme[ key ].strip().split(','):
```
Brobin: `strip()` doesn't remove spaces, only newlines and tabs.

`for value in theme[ key ].strip().replace(' ', '').split(','):`

rtv/curses_helpers.py
arnoutpierre: My idea was to allow us to define several themes in one file.

All files matching with `<RTV_CONFIG_PATH>/colors/*.cfg` are read because the theme id (example: `monokai-extended`) do not necessarily correspond to file name.

The goal is allow to create themes like:
```
[themename]
...

[themename-dark]
...

[themename-light]
...
```
This theme file will be surely name `themename.cfg`, but theme defined in `rtv.cfg` may be `themename-dark`

I don't know if it is a required feature, maybe we just should delete it.

Currently I read all themes and don't break the loop, at least I can break the loop when theme is found.

rtv/curses_helpers.py
arnoutpierre: Long time I have not do python, sorry

rtv/curses_helpers.py
Brobin: Oh, okay. This makes sense :+1:

rtv/__main__.py
```diff
@@ -24,7 +24,6 @@ def load_config():
     Search for a configuration file at the location ~/.rtv and attempt to load
     saved settings for things like the username and password.
     """
-
     config = configparser.ConfigParser()
```
michael-lazar: I would say rename this `config` to something else e.g.
```python
conf = configparser.ConfigParser()
```
That way you won't have to use a special import
```
from . import config as rtvconfig
```

rtv/config.py
```diff
@@ -1,5 +1,36 @@
 """
 Global configuration settings
 """
-
 unicode = False
+
+default_colors = {
```
michael-lazar: All of this extra whitespace goes against the standard python convention. See https://www.python.org/dev/peps/pep-0008/#pet-peeves for reference.

rtv/curses_helpers.py
```diff
@@ -183,9 +185,77 @@ def init(cls):
...
+        for key in theme:
+            values = []
+            for value in "".join( theme[ key ].split() ).split(','):
+                # check for aliases
+                if value in config.color_aliases.keys():
+                    values.append( config.color_aliases[ value ] )
+                else:
+                    # integer is required
+                    try:
+                        values.append( int( value ) )
```
michael-lazar: Whitespace, see comment above.

michael-lazar:
Hi! Thanks for the pull request, this looks awesome! I left some initial comments. I'm going to play with this over the weekend and see how it goes.

rtv/__main__.py
arnoutpierre: I chose `parser` for instance of `configparser.ConfigParser`, this is more accurate and avoid lot of `conf*` variables

arnoutpierre:
I've update the PR according to your comments :)

rtv/__main__.py
```diff
@@ -37,15 +36,20 @@ def load_config():
...
+        defaults = dict(parser.items('rtv'))
 
     if 'unicode' in defaults:
-        defaults['unicode'] = config.getboolean('rtv', 'unicode')
+        defaults['unicode'] = parser.getboolean('rtv', 'unicode')
+
+    if 'theme' in defaults:
+        config.theme = defaults['theme']
+    else:
+        config.theme = None
```
michael-lazar: FYI dictionaries have a `.get` method that does exactly what you're trying to do.

```python
config.theme = defaults.get('theme', None)
```
or simply
```python
config.theme = defaults.get('theme')
```

michael-lazar:
Hey I've got a few more comments now that I've had some time to try this out.

1. The number of color pairs that curses allows is dependant on which terminal you're using. I think it also has to do with the `TERM` environment variable ([reference](http://vim.wikia.com/wiki/256_colors_in_vim)). When I tried your theme out on my gnome-terminal I got the following error
```python
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/michael/Projects/rtv_project/rtv/__main__.py", line 139, in <module>
    sys.exit(main())
  File "/home/michael/Projects/rtv_project/rtv/__main__.py", line 116, in main
    with curses_session() as stdscr:
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "rtv/curses_helpers.py", line 375, in curses_session
    Color.init()
  File "rtv/curses_helpers.py", line 194, in init
    curses.init_pair(index, code[0], code[1])
_curses.error: init_pair() returned ERR
```
See https://docs.python.org/2/library/curses.html#curses.start_color. We need to at least support the following terminal configurations without crashing.
a.) Black and white
b.) 8 colors
c.) 256 colors
I'm not sure how exactly how `COLOR_PAIRS` is implemented, so there is some research required to figure out the best course of action to take.

2. The `theme` setting should be added as a runscript argument as well as a configuration value.
3. I think we should move the default theme into a `.cfg` file instead of defining it inside of the python code. This has the advantage of being easier to maintain (1 config format to parse instead of 2). In addition, we can use the default theme as a template for people to reference when creating their own themes. I noticed that your config was missing a couple of values, for instance the bars on the left side of the comments. I started drafting a more complete version and cleaned up the names a little. https://gist.github.com/michael-lazar/744bbd285f9da7d8f73f
4. We should add your monokai theme to the codebase, maybe in a seperate **themes** folder.
5. I like the way that you set up the `/colors` subdirectoy and having multiple themes per file. I think it's a good way to facilitate bunding and sharing themes. The only suggestion that I have is to maybe rename it to `/themes`.

I'll try to help out with these as much as I can, particularly #1. I'm working on adding unicode support ATM so I'm a bit tied up until I get that completed.

arnoutpierre:
Good, 

1. Like you said we have to investigate about this issue, I was not aware of that so I don't have nothing to telle about this issue right now.
2. I stashed some commits who allow us to define backgrounds and `levels` colors (the colors used for submission comments). But actually this feature add a new bug, I would like to have time to fix that today but I'm not sure. If you can help me to complete missing part it's will amazing. I'm ok for moving default theme in `.cfg` file
3. Yeah !
4. Ok for me

I will try to do this quickly !

LorenzLeitner:
Sorry for the unconstructive comment (there is no "like" button on pull requests), but I would also really like a theme plugin!

Brobin:
I like where this is headed! :plus1:

arnoutpierre:
Folks, I'm a bad guy. You know, this kind of guy who don't finalize his PR !
Sorry for that, I will delete this PR and do it again from sratch to avoid rebasing and conflicts.

I will probably try to clean some parts of my code too.

I will be back, soon

michael-lazar:
Ha no worries :stuck_out_tongue:. I look forward to seeing what you come up with for round 2.

arnoutpierre:
Hey @michael-lazar what's up ?

I soil this place before pushing a new branch. I have started a new one where I'm now at the same state as when this PR ended.

So now I'm thinking about our top issue for this PR: Terminal colors. (B&W / 8 / 256)

After a little reflection about that, I'm feeling the best solution would be to create a class / method that can take a color and translate it in other colors range.

Example:
```
class Color:
   def translate (cls, color, range):
      """ do stuff """
```
Usage:
```
# Translate a non-pure yellow to other ranges
Color.translate(230, "8")
# Retuns 11
Color.translate(230, "2")
# Returns 0
```

I'm thinking about that because any other solutions will force us to create a black & white and a 8 colors theme.

But I feel it's not a super sexy solution, just the "most" sexy of all I thinked
Do you have ideas, requirements, needs to explicit things you want / don't want ?

I don't like big dicts or lists in code, it's generally ugly as hell, but I do not found any algorithm that can translate terminal colors, so there is no other way to do it. 

Tell me what you think about all that please :)

PS: reference for color if needed: https://upload.wikimedia.org/wikipedia/en/1/15/Xterm_256color_chart.svg

michael-lazar:
Hey @arnoutpierre,

I can definitely see where you're coming from with this. I agree that translating the colors using some function would be better than maintaining a huge lookup table.

I don't know how I feel about about "downsampling" the colors though. I can picture someone coming up with a really clean 256 color scheme that just gets butchered when you try to convert it to 8 colors. On top of that, I remember reading that you can redefine your first 16 system colors to whatever you want. So it may come out even *worse* if we try to down-sample to those. At that point we would be better off raising a warning or even an error if the user can't load the colors in their config file. I'm curious what other terminal programs like vim do to handle this. Might be something to look into.

I think any way you slice it we're going to want to have at least a default for 8 colors. B&W is easier since you can just take the attributes (BOLD, REVERSED, etc.) from either the 8 or 256 color theme. Any thoughts on that? I like the idea of having one 8 color theme set as the gold standard and then having a couple of 256 themes bundled with the program.

arnoutpierre:
After a night to sleep I'm not really thrilled by translating colors. I don't like to store lot of thing, that's why I figured it will be a good idea to can translate colors, but this will be a pretty ugly method, I don't want to be the one who dirtied the code.

So, I tried to understand how Vim works:
```Note about colors: The 't_Co' option tells Vim the number of colors available.
When it is non-zero, the 't_AB' and 't_AF' options are used to set the color.
If one of these is not available, 't_Sb' and 't_Sf' are used.  't_me' is used
to reset to the default colors.```

What about that:.
```
# @color:  INTEGER
# @scheme: INTEGER (should be 0, 8 or 256)
def is_allowed(color, scheme):
    if color < scheme and color > -2:
        return True
    else:
        return False

# @color:  INTEGER
# @scheme: INTEGER (should be 0, 8 or 256)
# @type:   STR ("fg" / "bg"), tell us where to found default value if needed
def to_allowed(color, scheme, type = "fg"):
    if is_allowed(color, scheme):
        return color
    else:
        if type == "fg":
            return DEFAULT_FG
        else:
            return DEFAULT_BG

# How to process it
# . 0 Data used a default one (aka config.py)
DEFAULT_FG = 0
DEFAULT_BG = 0

# . 1 Init theme as dict
theme = {
    "color_range": 8,
    "default_bg": 10,
    "default_fg": 0,
    "color_1": 100,
    "color_2": 6,
    "color_3": 0,
}

# . 2 Get range
theme_range = int(theme["color_range"])
# verify things ..
if theme_range < 0 or theme_range > 256:
    theme_range = 0

# . 3 Verify default values
# This step is important because these values will be used as default one
# so we needs they exists
if not is_allowed(theme["default_fg"], theme_range):
    theme["default_fg"] = DEFAULT_FG
if not is_allowed(theme["default_bg"], theme_range):
    theme["default_bg"] = DEFAULT_BG

# . 4 Loop in other values to verify each color
# delete useless keys, yes it's dirty I know
theme.pop("color_range")
theme.pop("default_bg")
theme.pop("default_fg")

for color_key in theme:
    # the only bg color is currently "Background"
    # so for moment it may be possible to use static string
    # but we may want to at least create a static var to store it ..
    type = "bg" if color_key == "Background" else "fg"
    theme[color_key] = to_allowed(theme[color_key], theme_range, type)

print theme
```

michael-lazar:
I think this looks fine but it's hard to follow your code. If I'm understanding this correctly:

1. You set the color range inside of the theme definition. For example if the theme is meant to work with 8 colors you would let ``color_range: 8``. Good so far. :+1:
2. You give a default foreground and background color inside of the theme to replace any of the colors fall outside of the range. This is where I'm a little lost. Why would somebody write a theme for 8 colors and then add the ``100`` color a few lines down?

Issue #109, opened 2015-06-05
Support for receiving/replying to messages?


ghost:
Does the PRAW api support messages/inbox/etc? If so, would it be possible to implement a little notification when you have unread messages, a command to go to the inbox, reply to messages, things like that?

This is the only feature that keeps me opening reddit in a browser at this point.

I'd be happy to start working on this if it's something you think is feasible.

Brobin:
I'm fairly certain this is supported by praw. This would be great! Maybe it
could be added to the `/me` page.

mekhami:
What I was thinking was, when you use Reddit in the browser each request carries with it a notification if you have an unread message, so I assume that's part of the API. I'd like to be browsing normally in rtv and have a notification if I've received an inboxed message/comment/reply etc. 

So each call to PRAW would have to contain this notification. If it does, simple! If not, do we want some sort of timer that will hit the API and check for messages? How much can we do within each request to check for an inboxed item?

michael-lazar:
It's possible to check for unread messages with PRAW, but I don't believe that Reddit's API supports sending the notification with each request. You have to check for notifications explicitly (see [get_unread()](https://praw.readthedocs.org/en/v2.1.21/pages/code_overview.html#praw.__init__.PrivateMessagesMixin.get_unread)). The two approaches that you could take would be:

1. Implement a timer as you suggested. Have it make a new request every 30 seconds or so.
2. Check for messages every time the user switches to a new page or refreshes the screen.
3. Some hybrid of 1 and 2, where every time the user opens a new page it checks how much time has elapsed.

I would advise against option 1 because it will involve spinning off threads that need to interact with the main program loop. This would be a huge pain to implement. Option 2 could work, the main issue is that each PRAW request takes 2 seconds, and you don't want to block the user from doing anything for an additional 2 seconds every time they load a page. You might have to get creative here.

I think this is a cool idea. My suggestion would be to keep it simple at first and focus on just the notification. That in and of itself is worth adding. Once that is done, we can worry about viewing the inbox and replying to messages and such.

mekhami:
I wonder if you could fire off the two requests, one for get_unread() and
one for whatever the user was requesting, simultaneously. i would really
want the get_unread() call to be non blocking, or i don't think it would be
worth doing.

i'll do some research and get back to you on this. in the meantime if you
have implementation ideas i'm all ears.

mekhami:
[...]

PR #111, opened 2015-06-10
added the ability to check for unread messages with the 'i' key

mekhami:
Addresses Issue #109

Also two small PEP8 changes.

### DIFF
--- a/rtv/page.py
+++ b/rtv/page.py
@@ -417,10 +417,11 @@ class BasePage(object):
         """
         Checks the inbox for unread messages and displays a notification.
         """
+        inbox = len(list(self.reddit.get_unread(limit=1)))
         try:
-            if len(list(self.reddit.get_unread(limit=1))) > 0:
+            if inbox > 0:
                 show_notification(self.stdscr, ['New Messages'])
-            elif len(list(self.reddit.get_unread(limit=1))) == 0:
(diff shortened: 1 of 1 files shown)


rtv/page.py
```diff
@@ -412,8 +412,23 @@ def edit(self):
             s.catch = False
             self.refresh_content()
 
+    @BaseController.register('i')
+    def get_inbox(self):
+        """
+        Checks the inbox for unread messages and displays a notification.
+        """
+        try:
+            if len(list(self.reddit.get_unread(limit=1))) > 0:
```
michael-lazar: I think you need to move the praw call outside of the if-else statement, otherwise it may get called twice.

mekhami:
@michael-lazar addressed that, which seems so obvious in hindsight. It also works twice as fast now. who knew.

Brobin:
Is the plan to support actually viewing the inbox in the future? From what I see, this only checks if there is at least one new messages and shows it to you.

mekhami:
Yeah, I hope to add more functionality like that, but for now, i just want
to keep rolling out mini-features and iterating over it. I think it's a
good strategy here.

Brobin:
Cool. I was just clarifying because you said it addresses issue #109, but only does one part. Great work, I'm looking forward to this!

PR #116, opened 2015-07-07
Don't show flair when it's None

kbakulin:
praw could return None for \*flair\* params, don't show this.

Thats's what I saw before this fix:
rtv-01
and
rtv-02

### DIFF
--- a/rtv/submission.py
+++ b/rtv/submission.py
@@ -222,7 +222,8 @@ class SubmissionPage(BasePage):
         attr = curses.A_BOLD | Color.GREEN
         add_line(win, u'{author}'.format(**data), row, 1, attr)
         attr = curses.A_BOLD | Color.YELLOW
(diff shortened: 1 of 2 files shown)


michael-lazar:
That's interesting because we're doing this and it has historically worked.
```
flair = getattr(comment, 'author_flair_text', '')
```

Reddit must have changed from omitting the flair field to including it as a null value.

kbakulin:
I think more reliable way to do it is
```
flair = getattr(comment, 'author_flair_text', '') or ''
```
flair will be '' even if author_flair_text has None value. 
But then there will be printed one extra space, so I decided to use "if" checks.

michael-lazar:
Yea because of the empty space, at this point it might be better to do
```
flair = getattr(comment, 'author_flair_text', None)
```
and cut out the empty string all together. Either it exists (*valid string*) or it doesn't (*None*).

PR #126, opened 2015-07-30
Handle internal reddit links internally, including x-posts

obosob:
Thought it made sense to handle crossposts internally too. generally improved the determination of what is a self/cross post.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -163,7 +163,7 @@ class SubmissionContent(BaseContent):
 
         try:
             with loader():
-                submission = reddit.get_submission(url, comment_sort='hot')
+                submission = reddit.get_submission(url.replace("http:","https:"), comment_sort='hot')
         except praw.errors.APIException:
             raise SubmissionError(url)
 
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -107,7 +107,6 @@ class BaseContent(object):
         """
 
         reddit_link = re.compile("https?://(www\.)?(np\.)?redd(it\.com|\.it)/r/.*")
-        reddit_link_no_host = re.compile("/r/.*")
         author = getattr(sub, 'author', '[deleted]')
         name = getattr(author, 'name', '[deleted]')
         flair = getattr(sub, 'link_flair_text', '')
@@ -126,13 +125,18 @@ class BaseContent(object):
         data['flair'] = flair
         data['url_full'] = sub.url
 
-        if reddit_link.match(sub.url):
-            stripped_url = reddit_link_no_host.search(sub.url).group()
-            stripped_comments = reddit_link_no_host.search(sub.permalink).group()
-            data['url'] = ('selfpost' if stripped_url == stripped_comments
-                    else 'x-post')
+        if data['permalink'].split('/r/')[-1] == data['url_full'].split('/r/')[-1]:
+            data['url_type'] = 'selfpost'
+            data['url'] = 'selfpost'
+
+        elif reddit_link.match(data['url_full']):
+            data['url_type'] = 'x-post'
+            data['url'] = 'x-post via {}'.format(strip_subreddit_url(data['url_full']))
+
         else:
-            data['url'] = sub.url
+            data['url_type'] = 'external'
+            data['url'] = data['url_full']
+
         data['likes'] = sub.likes
         data['gold'] = sub.gilded > 0
         data['nsfw'] = sub.over_18
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -2,6 +2,7 @@ import logging
 
 import praw
 import requests
+import re
 
 from .exceptions import SubmissionError, SubredditError, AccountError
 from .helpers import humanize_timestamp, wrap_text, strip_subreddit_url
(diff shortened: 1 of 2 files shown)


rtv/subreddit.py
```diff
@@ -108,10 +108,11 @@ def open_link(self):
         "Open a link with the webbrowser"
         data = self.content.get(self.nav.absolute_index)
 
-        if data['url'] == 'selfpost':
-            self.open_submission()
+        url = data['url_full']
+        if data['url'] != url:
```
Brobin: This isn't very descriptive as to what's happening. If a cross post's url was represented as simply `'x-post'` instead of `'x-post via /r/{sub}'`, you could do something like this.

```python
if data['url'] in ['selfpost', 'x-post']:
    ...
```

I think this would make things more explicit, and obvious as to what's actually happening.

rtv/content.py
```diff
@@ -123,7 +125,14 @@ def strip_praw_submission(sub):
         data['subreddit'] = strip_subreddit_url(sub.permalink)
         data['flair'] = flair
         data['url_full'] = sub.url
-        data['url'] = ('selfpost' if is_selfpost(sub.url) else sub.url)
+
+        if reddit_link.match(sub.url):
+            stripped_url = reddit_link_no_host.search(sub.url).group()
+            stripped_comments = reddit_link_no_host.search(sub.permalink).group()
+            data['url'] = ('selfpost' if stripped_url == stripped_comments
+                    else 'x-post via {}'.format(strip_subreddit_url(sub.url)) )
```
Brobin: See previous comment. Would it be better to just have x-posts be `'x-post'`?

obosob:
I just felt it made a better view in the list if you know what sub its x-posted to. maybe i could add that as a special case in draw though. what do you think? I do prefer the readability of your suggestion.

Brobin:
Maybe we could add another key to the data. `data['x-post'] = '/r/python'`, then you could check that for display.

```python
if data['x-post']:
    # display "x-post via /r/sub"
```

Brobin:
Looks great! :+1:

rtv/content.py
```diff
@@ -105,7 +106,8 @@ def strip_praw_submission(sub):
         displayed through the terminal.
         """
 
-        is_selfpost = lambda s: s.startswith('http://www.reddit.com/r/')
+        reddit_link = re.compile("https?://(www.)?(np.)?redd(it.com|.it)/r/.*")
```
michael-lazar: I'm pretty sure your periods need to be surrounded with ``[.]``, otherwise they act like wildcards.

rtv/content.py
```diff
@@ -154,7 +163,7 @@ def from_url(cls, reddit, url, loader, indent_size=2, max_indent_level=8):
 
         try:
             with loader():
-                submission = reddit.get_submission(url, comment_sort='hot')
+                submission = reddit.get_submission(url.replace("http:","https:"), comment_sort='hot')
```
michael-lazar: Why is this necessary? If the URL is provided as ``http:``, it would make sense to honor that and let Reddit redirect us to the right location.

obosob: praw throws an unexpected redirect at this point otherwise. if you know of a better way to handle it then let me know.

rtv/content.py
obosob: damnit, you're right, well, escaping it will be fine, but yep.

rtv/content.py
```diff
@@ -123,7 +125,14 @@ def strip_praw_submission(sub):
...
         data['url_full'] = sub.url
-        data['url'] = ('selfpost' if is_selfpost(sub.url) else sub.url)
+
+        if reddit_link.match(sub.url):
+            stripped_url = reddit_link_no_host.search(sub.url).group()
+            stripped_comments = reddit_link_no_host.search(sub.permalink).group()
+            data['url'] = ('selfpost' if stripped_url == stripped_comments
+                    else 'x-post')
+        else:
+            data['url'] = sub.url
```
michael-lazar: I think that this the logic here is a kind of confusing to follow. I was thinking of ways to potentially clean it up and came up with this -

The values that we have are:
``data['permalink']`` - Link to the submission comments
``data['url_full']`` - Link that the submission points to
``data['url']`` - What is displayed on the screen

So in keeping with this convention, we should really add a new field for the post type. That way you don't have to hijack the ``data['url']`` for x-posts. I think Brobin was eluding to this also.

```python
if data['permalink'].split('/r/')[-1] == data['url_full'].split('/r/')[-1]:
    data['url_type'] = 'selfpost'
    data['url'] = 'selfpost'

elif reddit_link.match(data['url_full']):
    data['url_type'] = 'x-post'
    data['url'] = 'x-post via {}'.format(data['subreddit'])

else:
    data['url_type'] = 'external'
    data['url'] = data['url_full']
```

This also simplifies subreddit.py because you can go back to printing ``data['url']`` directly.

rtv/content.py
michael-lazar: Makes sense, thanks

rtv/content.py
Brobin: I think this is a great way to  handle this. Having `data['url_type']` instead of combinations of other conditions will make the code much more readable. :+1:

obosob: +1

obosob:
Bump...

michael-lazar:
Sorry, I must have missed your latest changes. Thanks for the PR. :smiley:

Issue #80, opened 2015-04-06
App crashes when trying to visit banned subreddits


Brobin:
I know this is a far-edge case, but I think the app should handle this.

Full trace:
```
Traceback (most recent call last):
  File "~/py3env/lib/python3.4/site-packages/rtv/content.py", line 240, in __init__
    self.get(0)
  File "~/py3env/lib/python3.4/site-packages/rtv/content.py", line 309, in get
    submission = next(self._submissions)
  File "~/py3env/lib/python3.4/site-packages/praw/__init__.py", line 508, in get_content
    page_data = self.request_json(url, params=params)
  File "~/py3env/lib/python3.4/site-packages/praw/decorators.py", line 163, in wrapped
    return_value = function(reddit_session, *args, **kwargs)
  File "~/py3env/lib/python3.4/site-packages/praw/__init__.py", line 561, in request_json
    retry_on_error=retry_on_error)
  File "~/py3env/lib/python3.4/site-packages/praw/__init__.py", line 403, in _request
    _raise_response_exceptions(response)
  File "~/py3env/lib/python3.4/site-packages/praw/internal.py", line 178, in _raise_response_exceptions
    response.raise_for_status()
  File "~/py3env/lib/python3.4/site-packages/requests/models.py", line 834, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/py3env/bin/rtv", line 7, in <module>
    from rtv.__main__ import main
  File "~/py3env/lib/python3.4/site-packages/rtv/__main__.py", line 133, in <module>
    sys.exit(main())
  File "~/py3env/lib/python3.4/site-packages/rtv/__main__.py", line 116, in main
    page.loop()
  File "~/py3env/lib/python3.4/site-packages/rtv/subreddit.py", line 45, in loop
    self.controller.trigger(cmd)
  File "~/py3env/lib/python3.4/site-packages/rtv/page.py", line 140, in trigger
    return func(self.instance, *args, **kwargs)
  File "~/py3env/lib/python3.4/site-packages/rtv/subreddit.py", line 88, in prompt_subreddit
    self.refresh_content(name=name)
  File "~/py3env/lib/python3.4/site-packages/rtv/subreddit.py", line 54, in refresh_content
    self.reddit, name, self.loader)
[...]

PR #128, opened 2015-08-06
Number keys to refresh the submission/comments and sort by hot/top/rising/new/controversial

yskmt:
I think it is related to:
https://github.com/michael-lazar/rtv/pull/11 and https://github.com/michael-lazar/rtv/pull/13

Keys:
1: hot, 2: top, 3: rising, 4: new, 5: controversial

The header shows the sorting method next to title

I also added `praw.errors.NotFound` in `SubredditContent.__init__` so this issue https://github.com/michael-lazar/rtv/issues/80 should be fixed too.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -137,7 +137,8 @@ class SubmissionContent(BaseContent):
     list for repeat access.
     """
 
-    def __init__(self, submission, loader, indent_size=2, max_indent_level=8):
+    def __init__(self, submission, loader, indent_size=2,
+                 max_indent_level=8, order='hot'):
 
         self.indent_size = indent_size
         self.max_indent_level = max_indent_level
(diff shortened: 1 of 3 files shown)


yskmt:
As I am submitting this PL, I realized that some subreddit has only a few sorting methods, and choosing the non-existing sorting method crashes the app (e.g. [/r/haskell](https://www.reddit.com/r/haskell)). That needs to be fixed...

michael-lazar:
Thanks, I'll look over it this weekend

michael-lazar:
Looks good! I went ahead refactored a few small things that I saw. I changed the header from ``python : new`` to ``python [new]`` because I think it looks a bit cleaner. I also hijacked this PR to fix a couple of bugs I found with the sorting order in general. Try it out when you get the chance and let me know if I broke anything :smile: 

As always thanks for the PR. :+1: We should add this to the documentation before the next release goes out.

yskmt:
Thanks, I didn't realize that I just have to edit `BasePage` class for common functionality :sweat_smile:

PR #129, opened 2015-08-07
Just a couple of little things

obosob:
add the intermediate setuptools files to the .gitignore and use the subreddit field from the PRAW Submission object instead of getting it out of the URL.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -122,7 +122,7 @@ class BaseContent(object):
         data['score'] = '{} pts'.format(sub.score)
         data['author'] = name
         data['permalink'] = sub.permalink
-        data['subreddit'] = strip_subreddit_url(sub.permalink)
+        data['subreddit'] = str(sub.subreddit)
         data['flair'] = flair
         data['url_full'] = sub.url
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .*
 *.pyc
-scripts/*
+scripts
+build
+dist
+rtv.egg-info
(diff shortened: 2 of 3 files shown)


rtv/content.py
```diff
@@ -122,7 +122,7 @@ def strip_praw_submission(sub):
         data['score'] = '{} pts'.format(sub.score)
         data['author'] = name
         data['permalink'] = sub.permalink
-        data['subreddit'] = strip_subreddit_url(sub.permalink)
+        data['subreddit'] = str(sub.subreddit)
```
michael-lazar: The last time I checked, calling the **sub.subreddit** property forced praw to make an extra API call to grab the subreddit object. This doubled the loading time because each call is rate limited to every 1 or 2 seconds. Can you verify that this is no longer the case?

obosob: OK, I've had a look into this and it's ok, the `Subreddit` object you get from `Submission.subreddit` is not "fetched" and has been initialised with the subreddit name and that is all that is returned from `Subreddit.__str__()`.

A lot of the calls on the `Subreddit` object *will* cause the cache to be refreshed (or initialised) but the subreddit name is not one of them. `__str__()` also does not cause the cache to be updated as it's just returning the name of the Sub, which is a param to `__init__()` passed to it by the `Submission` object.

In short, this is fine, but I understand your hesitance.

michael-lazar: Awesome, thanks for checking. I agree that your way is a lot nicer. :+1: praw is currently a *very* active project and sometimes it's hard to keep up with all of the changes going on over there.

Issue #104, opened 2015-05-23
subreddit jump list / tabs


gotbletu:
I dont like to waste time typing the subreddit everytime i want to switch.

Idea:
a way to bookmark all your favorite subreddit in the config.
Use a hotkey to see the list, then jump to the subreddit you want quickly.
Or add tabs support

michael-lazar:
I've thought about this a bit myself as well. I think the best thing to do would be to display the user's subscribed subreddits. AFAIK this is Reddit's equivalent to "bookmarking". We can grab the list with praw's [get_my_subreddits](https://praw.readthedocs.org/en/v2.1.21/pages/code_overview.html#praw.__init__.MySubredditsMixin.get_my_subreddits) function. The tricky part will be displaying the list with curses and supporting stable scrolling. I'm all for it though! There's a large backlog of issues that I'm currently working through ATM. We're very open to contributions if you want to try putting something together.

tobywhughes:
I could see something along the lines of when entering in the subreddit name, have a button that opens up a prompt that allows you to select your subscribed subreddits. If there are more than a certain amount, have a a key like [ and ] to select a different page. I don't really know much about curses, but I'll see if I can put something together.

Oh, and hi everybody! Haven't been able to contribute for a while cause of school and work, but just finished finals so I'll be around for a couple of weeks.

michael-lazar:
Closing, as a similar feature is available through opening the user's subscriptions

PR #131, opened 2015-08-09
Adding "subscriptions" feature

TheoPib:
Added the ability for authenticated users to view their subscribed subreddits, and to jump to one of them.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -153,7 +153,7 @@ class BaseContent(object):
         data = {}
         data['object'] = subscription
         data['type'] = 'Subscription'
-        data['name'] = subscription._case_name
+        data['name'] = "/r/" + subscription._case_name
         data['title'] = subscription.title
 
         return data
@@ -404,9 +404,8 @@ class SubscriptionContent(BaseContent):
                 self._subscription_data.append(data)
 
         data = self._subscription_data[index]
-        subreddit_info = "/r/" + data['name'] + "\n" + data['title']
-        data['split_title'] = wrap_text(subreddit_info, width=n_cols)
-        data['n_rows'] = len(data['split_title'])
+        data['split_title'] = wrap_text(data['name'], width=n_cols)
+        data['n_rows'] = len(data['split_title']) + 1
         data['offset'] = 0
 
         return data
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -368,22 +368,21 @@ class SubredditContent(BaseContent):
         return data
 
 class SubscriptionContent(BaseContent):
-    def __init__(self, name, subscriptions, loader):
-        self.name = name
+    def __init__(self, subscriptions, loader):
+        self.name = "Subscriptions"
         self._loader = loader
         self._subscriptions = subscriptions
         self._subscription_data = []
 
     @classmethod
     def get_list(cls, reddit, loader):
-        display_name = "Subscriptions"
         try:
             with loader():
                 subscriptions = reddit.get_my_subreddits(limit=None)
         except praw.errors.APIException:
             raise SubscriptionError()
 
-        return cls(display_name, subscriptions, loader)
+        return cls(subscriptions, loader)
 
     def get(self, index, n_cols=70):
         """
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -404,7 +404,7 @@ class SubscriptionContent(BaseContent):
                 self._subscription_data.append(data)
 
         data = self._subscription_data[index]
-        subreddit_info = "/r/" + data['name'] + " - " + data['title']
+        subreddit_info = "/r/" + data['name'] + "\n" + data['title']
         data['split_title'] = wrap_text(subreddit_info, width=n_cols)
         data['n_rows'] = len(data['split_title'])
         data['offset'] = 0
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -368,21 +368,22 @@ class SubredditContent(BaseContent):
         return data
 
 class SubscriptionContent(BaseContent):
-    def __init__(self, subscriptions, loader):
-        self.name = "Subscriptions"
+    def __init__(self, name, subscriptions, loader):
+        self.name = name
         self._loader = loader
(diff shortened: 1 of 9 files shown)


michael-lazar:
Hi Theo. Thanks for the PR, this is very impressive! This is a large addition so I will probably need a few days to go over everything in more detail. At first glance the code looks clean and well though out. I'll get back to you if I have any comments or questions.

michael-lazar:
Relates to https://github.com/michael-lazar/rtv/issues/104

TheoPib:
I tried to stick with rtv's code structure, maybe some portions of the code can be improved, but the core feature is a call to PRAW's get_my_subreddits().

TheoPib:
I found out that pressing s key while in subscriptions screen would make rtv crash due to a missing import, but instead of unnecessarily importing SubscriptionPage, I preferred overriding 's' mapping in SubscriptionController.

TheoPib:
Well, I just noticed I left out the initial draft of get_subscriptions() method in page.py. I removed it since it is now in SubredditController, which by the way fixes the bug I mentioned above. I however kept the updated keybinding.

rtv/content.py
```diff
@@ -143,6 +143,20 @@ def strip_praw_submission(sub):
...
+    def strip_praw_subscription(subscription):
+        """
+        Parse through a subscription and return a dict with data ready to be
+        displayed through the terminal.
+        """
+
+        data = {}
+        data['object'] = subscription
+        data['type'] = 'Subscription'
+        data['name'] = "/r/" + subscription._case_name
```
michael-lazar: Any reason we can't use ``subscription.display_name``? I would rather avoid private variables if possible.

rtv/subscriptions.py
```diff
@@ -0,0 +1,72 @@
...
+        self.content = SubscriptionContent.get_list(self.reddit, self.loader)
+        self.nav = Navigator(self.content.get)
+
+    @SubscriptionController.register(curses.KEY_ENTER, 10, curses.KEY_RIGHT)
+    def open_selected_subreddit(self):
+        "Open the selected subreddit"
+
+        from .subreddit import SubredditPage
+        data = self.content.get(self.nav.absolute_index)
+        page = SubredditPage(self.stdscr, self.reddit, data['name'][2:]) # Strip the leading /r
```
michael-lazar: You shouldn't need to strip the `/r/`, SubredditContent.from_name() does that for you!

rtv/subscriptions.py
```diff
@@ -0,0 +1,72 @@
...
+        "Re-download all subscriptions and reset the page index"
+
+        self.content = SubscriptionContent.get_list(self.reddit, self.loader)
+        self.nav = Navigator(self.content.get)
+
+    @SubscriptionController.register(curses.KEY_ENTER, 10, curses.KEY_RIGHT)
+    def open_selected_subreddit(self):
+        "Open the selected subreddit"
+
+        from .subreddit import SubredditPage
```
michael-lazar: Is this because of circular imports? If so, please put a comment here saying so. I would be nice if there was a better way to handle this :/

rtv/content.py
```diff
@@ -352,3 +366,46 @@ def get(self, index, n_cols=70):
...
+                with self._loader():
+                    subscription = next(self._subscriptions)
+            except StopIteration:
+                raise IndexError
+            else:
+                data = self.strip_praw_subscription(subscription)
+                self._subscription_data.append(data)
+
+        data = self._subscription_data[index]
+        data['split_title'] = wrap_text(data['name'], width=n_cols)
```
michael-lazar: So I think in this case, ``subreddit.title`` (data['title']) is actually the one that should be wrapped and not the name. The title is the property with the subreddit description and may be longer than the width of the terminal.

rtv/subscriptions.py
```diff
@@ -0,0 +1,72 @@
...
+    @staticmethod
+    def draw_item(win, data, inverted=False):
+        n_rows, n_cols = win.getmaxyx()
+        n_cols -= 1  # Leave space for the cursor in the first column
+
+        # Handle the case where the window is not large enough to fit the data.
+        valid_rows = range(0, n_rows)
+        offset = 0 if not inverted else -(data['n_rows'] - n_rows)
+
+        n_title = len(data['split_title'])
```
michael-lazar: Based on my comment in ``content.py``, you would have to switch this around to add ``data['name']`` first and then ``data['split_title']``.

michael-lazar:
Hey sorry for the delay, I left some comments!

TheoPib:
I'm going to push a commit to push the fixes you mentioned. As for the circular imports, I know my solution is not ideal, but I was unable to find a better way unfortunately. I'll try again though.

rtv/content.py
TheoPib: No reason at all, I simply didn't find a better variable at this time.

TheoPib:
Well, I'm excited to say that I found a way to :
- Avoid circular imports or import in a function
- Improve the way a selected subreddit in subscriptions list is opened : instead of creating a new instance of SubredditPage, I store the selected subreddit data in an object, set active to False, I get the data from the caller SubredditPage instance, and finally I refresh with the selected subreddit.

This approach seems better overall (no need to recreate a SubredditPage when there is already one in the back).
I also addressed the other points you mentioned.

michael-lazar:
Nicely done, I really like that approach! I went ahead and merged this with develop. There were three things not worth making another round of comments so I just changed them myself:

- Got rid of the getter method and opted for grabbing ``page.selected_subreddit_data`` directly.
- Added the vim binding ``h`` to exit the subscription page
- Switched back to having the subreddit name on top with the description underneath . This was my fault for poorly communicating what I was asking for.

Thanks again I think this turned out really well.

TheoPib:
No problem, I'm really happy to contribute to your project!

I saw the additional changes you made to the code. When I modified the wrapped field, I was unable to leave the subreddit name on top. I guess I didn't fully understand the way it was done, but you fixed it, so it's alright now.

PR #132, opened 2015-08-10
Don't forget to add selfposts and x-posts to history

obosob:
Links for selfposts and x-posts don't go purple after my last PR (#126). This'll fix it.

### DIFF
--- a/rtv/subreddit.py
+++ b/rtv/subreddit.py
@@ -106,23 +106,19 @@ class SubredditPage(BasePage):
         page = SubmissionPage(self.stdscr, self.reddit, url=data['permalink'])
         page.loop()
 
-        if data['url'] == 'selfpost':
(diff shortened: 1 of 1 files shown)


rtv/subreddit.py
```diff
@@ -106,23 +106,19 @@ def open_submission(self):
         page = SubmissionPage(self.stdscr, self.reddit, url=data['permalink'])
         page.loop()
 
-        if data['url'] == 'selfpost':
```
michael-lazar: Don't forget this function also gets called when you press the right or "l" key, we want to add to the history in that case too

obosob: Ah, yeah, you're right.

michael-lazar:
Nice catch, thanks!

PR #133, opened 2015-08-17
OAuth

TheoPib:
Hi, I saw OAuth feature in the backlog, so I tried to implement OAuth-based RTV.
Since this is a huge change, I'll try to explain how I did.

First, I created a "Reddit Terminal Viewer" app in Reddit (as Installed app). This app type allows authentication without client secret (PRAW however still complains if no client secret is given, so I just put some text in it, since it is not used.)

Then, since Reddit needs a callback URI where the data will be transmitted via GET method, I created a small Flask webapp to serve as a callback. I hosted it on my [server](https://rtv.theo-piboubes.fr) (SSL certificate is CAcert, which may not be present in database, thus possibly raising an error. It's still signed however). I also pushed the webapp on my [GitHub](https://github.com/TheoPib/oauth_helper). It does not store anything, it justs displays data provided by Reddit.

Finally, I implemented the OAuth logic on Reddit Terminal Viewer, which works this way:

If no previous OAuth data is stored in the configuration file, it begins the authentication procedure :
* Opens a webpage asking for access (managed by Reddit). The scopes may have to be reviewed.
* Once user allowed access, he is redirected to OAuth Helper (see above), displaying two strings :
    * **State** (verify that the user is the one making the request)
    * **Code** (random string that will be exchanged for an access token).
They must be entered in the two in-app prompts.
* Finally, an access token is retrieved from Reddit servers, and the user is logged in. The app manages token refresh.

If previous OAuth data is found, it just refreshes the access token, and proceeds.

It is possible to change the Reddit app settings in config.py (client ID, redirect URI), so that one can use his own app and server if he doesn't trust the current setup.

I hope I'm clear enough on the way it works.
Furthermore, I made important changes to the app itself. I hope the changes I did are okay, I still can modify if needed.


michael-lazar:
Hey thanks! I've been meaning to tackle this for a while. I've made a few attempts at it but never got to the point where I was satisfied. I'm really paranoid about breaking the user experience. I don't like the idea of having to open a web browser to login, kinda defeats the whole purpose of having a terminal application. Ideally I think things are best the way they are now, but I understand that the change is inevitable.

A few thoughts:
- I agree that installed app is the way to go. Have you looked at the [Implicit grant flow](https://github.com/reddit/reddit/wiki/OAuth2#authorization-implicit-grant-flow)? It looks simpler but it doesn't support permanent tokens and isn't supported by PRAW yet :/ Any thoughts? I'm still trying to wrap my head around how all of this works.
- I really dislike having to rely on a third-party web server for validation. Not that I don't trust your server (I do) but there are reliability and security concerns that people will inevitably bring up. Have you thought about just launching the web server from inside of RTV and setting the callback to http://127.0.0.1:65015 or something like that? There are a couple of smaller praw packages, https://github.com/avinassh/prawoauth2 and https://github.com/SmBe19/praw-OAuth2Util that do it this way. Another advantage is that you can grab the *state* and *code* directly without having people copy them manually.
- To take that a step further, I've though about prompting the username/password on the terminal and using **requests** to simulate clicking the *allow authentication* button. In this case you could inspect the redirect query params directly and wouldn't need a webserver at all! You could even show a little popup in the terminal with the scope permissions and an **accept** button. This would be my preferred way of sticking it to OAuth and maintaining the current flow. The downside is that it would be pretty hacky to implement and might go against reddit TOS. What do you think? I might try it one of these days even if we go ahead with a different approach.
- Have you thought about how this will handle logging out and logging in as a different user? Would it always store the most recent user in the config?

TheoPib:
I've read your thoughts, and there are some points I already thought of.

* I've been playing with Reddit's **implicit grant** flow. This one makes the OAuth process implementation easier, but as you said, it doesn't support permanent tokens, which is a no-go to me. This brings up the "web browser" issue. I understand that you dislike having to use a web browser, I've been thinking about this as well. I found that using the **code** flow kind of limits the damage, since the web browser is used only once. Implicit grant flow would require web browser each time the token expires, that is every hour. So my opinion is that the **code** flow is the best option.

* I agree with you about the third-party web server. In fact, I did set it up for myself, and I'm not sure it is a good idea to deploy OAuth in the main RTV with my webserver. In this regard, I tried to make it easy to edit the OAuth app settings (in config.py). On the other hand, using a stand-alone webserver could be a security issue, as it is not using HTTPS, unencrypted OAuth data would transit through Reddit servers and user's PC. So unless we configure SSL on the stand-alone server, we have to keep that mind as well.

* Providing a small interface to Reddit's authorization page could be a possibility. I haven't thought of this yet, we could discuss this one. But as you said, it's quite hacky, and to me it feels like it will complicate things. If you ask me, I prefer having to click "Accept" on a web browser once, but this is only my opinion.

* I haven't thought of that either. Now in the current setup, logging in as a different user would need one to logout from his web browser, login as the other user, and redo the procedure. Now this is indeed not ideal. Maybe asking for username and password would be a good idea.

michael-lazar:
The webserver would only be for the redirect. So correct me if I'm wrong, but it would look like this

```
----------------
|www.reddit.com|
----------------
       |
       | HTTPS
       |
       V
---------------------------------------
|    localhost                        |
| ---------------  HTTP  ------------ |
| |  web browser| ------>| Flask app| |
| ---------------        ------------ |
---------------------------------------
```
As long as the HTTP doesn't extend out of localhost I think we're good.

TheoPib:
If it sends the OAuth data to the client via HTTPS which then goes to the redirect URI, then it's OK. Still not convinced about it though, I should try to find an explanation of the process in reddit docs...

So if it works as you described, then yes we can do a standalone Flask app.

michael-lazar:
Other than that I agree with your points, thanks for the feedback
- **implicit grant** using the **code** workflow
- Stick to opening the web browser for the user to authenticate. Maybe we can revisit this one at a later point in time.

TheoPib:
I agree with the fact that the redirect URI should be localhost. I'm not too confident about having people reach my server, as it is more for personal use.

TheoPib:
I looked at the OAuth reddit wiki page, and the process we discussed above is not explained directly, but I've seen this :

### Allowing the user to authorize your application

If the user chooses to allow your application, their browser will be instructed to redirect to your app's registered `redirect_uri`.

So it seems that you're right, it is only a redirection, all the data goes through HTTPS between Reddit and the client.

I'll bundle the webserver in RTV then.

TheoPib:
Well, I've finished integrating the webserver into RTV. I chose Tornado in the end, since it has a better IO control (able to stop the execution as soon as the state and code are obtained) than Flask AFAIK. There are some ResourceWarning about unclosed files and sockets. I tried to fix it, but this is a Tornado problem, so there was not much I could do about it. I don't know if it is a concern to you ?

Otherwise, integrating the web server was a great idea. It significantly reduces user input (OAuth process is almost automatic now, the user only have to close the tab where the callback page has been opened), which makes the process seamless.

If you have any questions, I'm available.

TheoPib:
Actually, there was something I could do about ResourceWarnings. I got a bit mislead by Tornado's documentation.

michael-lazar:
Thanks for the update Theo. I think Tornado is a solid choice. I've been busy lately and haven't had a chance to go over the latest changes, but I promise I'll get to it soon! 

I'm also currently working on cleaning up the config handling in anticipation for custom commands and color themes. I'm thinking if I can get that cranked out by the end of the week it would be good to bundle that with these OAuth changes into a single release.

michael-lazar:
Hey so I finally got the chance to play around with this branch for a while. Overall the flow is very nice, better than I ever expected oauth to be! 

I'm trying something new and creating a milestone for this feature. I've linked separate issues for the remaining tasks that I found while testing. I also copied this branch onto the main repository under the name *oauth*. If you want to work on these issues, you're more than welcome to make the changes directly on this PR or by making new PRs off of the *oauth* branch. If you're burnt-out on this because I took too long to respond (understandable!), let me know and I'll try to start cranking through them.

PR #135, opened 2015-08-24
Added config option to get password by running external command

ellisadigvom:
Because keeping passwords in plain text gives me the creeps.


michael-lazar:
Hi Ellis, thanks for the contribution!

I understand your concern and other people have brought this up in the past as well. The current suggested flow is to leave the password blank and just type it into the prompt every time you launch the application. What external command are you using? Is there any reason you can't do ``rtv -p $"{command}"``?

ellisadigvom:
There's no reason, actually. I simply thought it would be more convenient to have all the configuration in one place. Though now that I think of it, I can accomplish that with an alias.

Nevermind then.

PR #136, opened 2015-08-27
add undocumented function to display new messages

shaggytwodope:


michael-lazar:
Hey thanks! I think this should be in the authenticated commands section (doesn't work unless you're logged in)

shaggytwodope:
@michael-lazar Adjusted. Didnt cross my mind to put in that section lol.

PR #142, opened 2015-09-02
Terminal web browser support (OAuth)

TheoPib:
I implemented terminal web browser support for OAuth. I've tested it with elinks and it worked (the browser must however be closed manually after allowing access for the authentication flow to continue). It seems to be working with the tests I did, but they may not be sufficient. Keep me updated on possible issues.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,9 +163,10 @@ RTV will read two configuration files:
 Each line in the files will replace the corresponding default argument in the launch script.
 This can be used to avoid having to re-enter login credentials every time the program is launched.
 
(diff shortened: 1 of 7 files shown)


TheoPib:
Also, since OAuthTool and therefore Tornado web server are started only if needed (initial authentication), the "Address already is use" exception that would appear when running two RTV simultaneously (caused by two concurrent Tornado instances trying to listen on the same port) is gone.

TheoPib:
Actually, I'm not satisfied with the way it currently works. I feel I'm making things more complicated than they should be. Closing this and starting over.

PR #143, opened 2015-09-04
OAuth terminal web browser support

TheoPib:
Hey, I finished implementing terminal web browser support. It was much easier with ``curses.endwin()`` and ``curses.doupdate()``. I had looked Python Curses documentation, but I guess I couldn't figure ``curses.doupdate()`` would set screen back up, maybe because the documentation was not very explicit about it. Anyway, it is working now, and much cleaner than my previous attempt. :)

### DIFF
--- a/rtv/oauth.py
+++ b/rtv/oauth.py
@@ -28,6 +28,9 @@ class HomeHandler(web.RequestHandler):
 
 class AuthHandler(web.RequestHandler):
 
+    def initialize(self):
+        self.compact = os.environ.get('BROWSER') in ['w3m', 'links', 'elinks', 'lynx']
+
     def get(self):
         global oauth_state
         global oauth_code
@@ -38,6 +41,10 @@ class AuthHandler(web.RequestHandler):
         oauth_error = self.get_argument('error', default='error_placeholder')
 
         self.render('auth.html', state=oauth_state, code=oauth_code, error=oauth_error)
+
+        # Stop IOLoop if using BackgroundBrowser (or GUI browser)
+        if not self.compact:
+            ioloop.IOLoop.current().stop()
 
 class OAuthTool(object):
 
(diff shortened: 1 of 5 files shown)


rtv/oauth.py
```diff
@@ -99,15 +110,25 @@ def authorize(self):
...
             permission_ask_page_link = self.reddit.get_authorize_url(str(hex_uuid),
                 scope=self.scope, refreshable=True)
 
-            with self.loader(message='Waiting for authorization'):
-                webbrowser.open(permission_ask_page_link)
+            if self.compact:
+                show_notification(self.stdscr, ['Opening ' + os.environ.get('BROWSER')])
+                curses.endwin()
+                webbrowser.open_new_tab(permission_ask_page_link)
                 ioloop.IOLoop.current().start()
```
michael-lazar: Have you tested this out? I still get stuck waiting for a reply from ``127.0.0.1:65000``. I assume this is because the iloop isn't started until after the webbrowser closes, but the webbrowser won't redirect unless the iloop has already started. I haven't used tornado before but it looks like there's some anync stuff built-in (``iloop.run_sync()``) that may be useful.

TheoPib:
The redirection is never done with terminal web browser, but if you close the web browser after clicking Authorize, the process goes on normally. Now it may be possible to make the redirection work, but the user will still have to close the web browser after landing on the page, since the browser runs in foreground. I'll see what I can do.

rtv/oauth.py
TheoPib: The redirection is never done with terminal web browser, but if you close the web browser after clicking Authorize, the process goes on normally. Now it may be possible to make the redirection work, but the user will still have to close the web browser after landing on the page, since the browser runs in foreground. I'll see what I can do.

TheoPib:
With the three most recent commits, the web browser no longer waits for 127.0.0.1, it goes to the callback page, where the user is instructed to close the page (or the web browser). Also, HTTP server is started only once to avoid "Address already in use".

TheoPib:
I added concurrent.futures backport requirement if user runs Python 2, since it was added in Python 3.2.

PR #146, opened 2015-09-20
added a manpage

shaggytwodope:
Added a simple man page, if this is a shit idea. Or you'd like something re-done let me know. 

Most of it is taken from -h output, be glad to redo any bit.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,4 @@
 include version.py
 include CHANGELOG.rst CONTRIBUTORS.rst LICENSE
-<<<<<<< HEAD
 include rtv.1
-=======
 include rtv/templates/*
->>>>>>> 28d17b28d0840f75386586686897e9316378150e
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include version.py
 include CHANGELOG.rst CONTRIBUTORS.rst LICENSE
+include rtv.1
--- a/rtv.1
+++ b/rtv.1
@@ -1,0 +1,113 @@
+.TH "rtv" "1" "September 2015" "Version 20150919" "Usage and Commands"
+.SH NAME
+rtv - Browse Reddit from your terminal
+.SH SYNOPSIS
+rtv [-h] [-s SUBREDDIT] [-l LINK] [--ascii] [--log FILE] [-u USERNAME]
+.SH DESCRIPTION
+.B Reddit Terminal Viewer
+is a lightweight browser for www.reddit.com built into a
+terminal window.
+.SH OPTIONS
+.TP
+\fB\-h\fR, \fB\-\-help\fR
+show this help message and exit
+.TP
+\fB\-s\fR SUBREDDIT
+subreddit name
+.TP
+\fB\-l\fR LINK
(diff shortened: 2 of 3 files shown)


michael-lazar:
Hey this is great! I'm not too familiar with man pages, is there a guide/example that you went off of when setting up the **setup.py** file? I would love to take a look out of curiosity. Also does this work on both linux and osx?

shaggytwodope:
Python has no native mechanism I'am aware of to handle manpages, Alot of projects simply use the data_files feature in setup.py to do this. I basicly searched all over the web on how to do this. 

One example would be https://github.com/0x90/scapy-osx/blob/master/setup.py

I see no reason not to expect it to work for osx (alot of stuff I googled does this the same). However I've no such hardware to test on, so can't confirm this on my own.

shaggytwodope:
Little bit of googling http://serverfault.com/questions/391834/where-are-os-x-man-pages-stored
Assuming no odd env settings by the user, this should work. However as I mentioned I've no way to confirm. If anyone can confirm this tho as the default path that'd be wonderful.

michael-lazar:
Yea I just checked [youtube-dl](https://github.com/rg3/youtube-dl/blob/master/setup.py) and they do the same thing as well (share/man/man1). It looks like it should work for at least the majority of distros.

michael-lazar:
I'm thinking about removing the "Controls" section. There are quite a few keys and I foresee it becoming a pain to maintain the definitions across 3 different files (readme, docs.py, and now the man page). I think it would be ok to add something like "type ? for a list of commands" to the description section instead. Thoughts?

Also how about adding an ENVIRONMENT section with short descriptions for $RTV_EDITOR and $BROWSER, since those are both used environment variables that we use.

michael-lazar:
Also a FILES section that lists the config file location

shaggytwodope:
https://u.teknik.io/KcYxNP.png
Added the sections you mentioned, tossed a screenshot cause I can. As far as removing the controls part just say the word and I can remove it. Just mention the basics maybe, and then point towards the feature to press ? and view list in rtv. I think thats a nice balance, a man page doesn't need to be utterly details if the program offers some basics for use imo.

As always anything needs to be fixed or something let me know, I'am far from the greatest editor ever lol.

michael-lazar:
Good point. I would say go for it, remove the controls section and replace with a short blurb (a sentence or two) about the basic controls and pressing ``?``. You can get rid of the authentication section too, the master branch has switched to using only OAuth now.

shaggytwodope:
Alrighty, edited accordingly.

michael-lazar:
Merged! I made some minor changes in style, I hope you don't mind.

I also had some fun and turned the manpage into a template. So now when you do something like change the version number or add a new command line argument you can simply run ``python scripts/build_manpage`` and the page will be rebuilt. That way, hopefully it won't fall out-of-date with the rest of the documentation!

shaggytwodope:
No worries at all on any changes. I need to remember to add the new manpage generation to the new release version for the AUR package.

michael-lazar:
As of now you shouldn't have to do anything, I'll run it everytime there's a new release and commit the updated manpage to the master branch.

PR #151, opened 2015-10-02
Make loads interruptible and some more small things.

Ferada:
Hi, this adds one feature I'm missing, the ability to cancel loads (because bad network etc.). It's similar to the `safe_call` mechanism, except it responds to <kbd>Ctrl-C</kbd>/`SIGINT` during blocking loads (while "Downloading..." shows). I haven't *yet* annotated all instances of these (mostly because I'd have to find them first), but the initial download, the comment thread and the "more comments" fields have been changed.

The other commits are more cleanups, although I noticed that `curses.wrapper` was initially used, but then rewritten into `curses_session`; I'm not sure why, so I included this proposal to reduce the lines of code again.

I'm happy to restructure this if doesn't look good so far, but I also want early feedback if this is even wanted. Btw. is there a common location for chat/discussion so far, except in the issues tracker?

### DIFF
--- a/rtv/__main__.py
+++ b/rtv/__main__.py
@@ -63,7 +63,8 @@ def main():
         print('Connecting...')
         reddit = praw.Reddit(user_agent=AGENT.format(version=__version__))
         reddit.config.decode_html_entities = False
-        with curses_session() as stdscr:
+
+        def loop(stdscr):
             oauth = OAuthTool(reddit, stdscr, LoadScreen(stdscr))
             if oauth.refresh_token:
                 oauth.authorize()
@@ -71,8 +72,11 @@ def main():
             if args.link:
                 page = SubmissionPage(stdscr, reddit, oauth, url=args.link)
                 page.loop()
+
             page = SubredditPage(stdscr, reddit, oauth, args.subreddit)
             page.loop()
+
+        curses_session(loop)
     except (praw.errors.OAuthAppRequired, praw.errors.OAuthInvalidToken):
         print('Invalid OAuth data')
     except praw.errors.NotFound:
--- a/rtv/__main__.py
+++ b/rtv/__main__.py
@@ -2,6 +2,7 @@ import os
 import sys
 import locale
 import logging
+import signal
 
 import requests
 import praw
@@ -10,7 +11,7 @@ import tornado
 
 from . import config
 from .exceptions import SubmissionError, SubredditError, SubscriptionError, ProgramError
-from .curses_helpers import curses_session, LoadScreen
+from .curses_helpers import curses_session, KeyboardInterruptible, LoadScreen
 from .submission import SubmissionPage
 from .subreddit import SubredditPage
 from .docs import AGENT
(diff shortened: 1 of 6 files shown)


michael-lazar:
Hi, thanks for the PR! I looked over it briefly and I think most of the clean-up code looks good ;) I would say don't go crazy on refactoring but if you notice simple things like the configuration loading changes you made, those are fair game. 

It's still a small project so the issue tracker is your best bet for communication. I prefer that to email because it leaves a public record and other have a chance to join in the conversation. Feel free to post general discussion or questions there too, it doesn't have to be limited to specific issues.

A couple comments on the KeyboardInterruptible
- First off, I really like the idea of cancelling loads and I think it's a valuable feature to add.
- Why are you altering SIGINT? I'm not too familiar with it but from what I understand  (http://stackoverflow.com/a/1112350) if you send a signal to kill the program, it should kill the program. If I type ``kill rtv`` the whole thing should exit immediately no matter what is happening inside.
- Would it make sense to add this functionality directly into the LoadScreen class? That would theoretically cover every time we could want to cancel something from loading.
- Continuing on the previous points, to me it feels more intuitive to use the ``esc`` key to cancel. Escape is already used in other places in the code to cancel operations, and there's an EscapeInterrupt exception that you can raise. Would be a bit harder to implement (maybe add a stdscr.getch() to the loading-screen's background thread that raises an exception in when esc is pressed?). It just feels dirty to override ctrl-c to mean something other than "terminate".

michael-lazar:
Also please excuse me, I will be on vacation this week and may be slow/absent to respond until I get back.

Ferada:
- Regarding altering `SIGINT`, that signal is usually send with <kbd>Ctrl-C</kbd>, but it's different from `SIGQUIT` or `SIGKILL`, so it will *only* affect this specific use case; killing the process with `kill` uses `SIGTERM` by default, so that's already different and the alteration explicitly doesn't handle `SIGTERM`.
- Regarding adding it to `LoadScreen`, possibly. I didn't want to make it too complicated, but I'll look if that works with minimal alterations, i.e. the same protocol with `.disable()` (respectively `done = True`).
- It would indeed be better to cancel on escape/any other customisable key. However as far as I understood the loading happens synchronously, so there's no way to interrupt it except via signals? I'll look again at the background thread, but that's the reasoning why I choose to do it this way.

michael-lazar:
Gotcha, thanks for clearing the ``SIGINT`` up.

michael-lazar:
Hey! Over the past few I've been working to refactor parts of the project and I wanted to give an update. 

First of all I apologize for not merging this PR, it came at a busy time for me and kind of fell between the cracks. However I was able to draw a lot of inspiration from it. I ended up going forward and adding this functionality into the ``loader``'s context manager. I also increased the usage of ``loader`` throughout to code to cover almost every blocking HTTP call. As a result, the program as a whole is a lot more responsive now and has straightforward error handling when something doesn't load. Also, I was able to hook the <kbd>Esc</kbd> key to trigger a SIGINT from the ``loader``'s thread, so now you can cancel a request without having to press <kbd>ctrl-c</kbd>

Ferada:
Great, thanks for the update.

Issue #153, opened 2015-10-07
Change URL field of self posts


zambonin:
Hello,

Do you think adding `self.subreddit_name` instead of `selfpost` to the `data['url']` field would be a welcome addition so to mirror the information displayed on the original site, or would it be classified as duplicate content, and hence, useless to the layout of the application?

I do believe that this is managed on [this](https://github.com/michael-lazar/rtv/blob/master/rtv/content.py#L136) line, and it wouldn't be too big of a change.

Thanks.

[edit] Other cosmetic features such as `/u/` before the username to match the `/r/` before the subreddit, separate the time and the comment counts with a black dot, bracketed tags etc. also could be changed in a whim.

michael-lazar:
Hi, apologies for the delayed response. 

There are two main goals that keep in mind when making formatting changes:

1. The layout should be compact so it can fit on small terminal screens with minimal text wrapping or ugliness.
2. The sections should be easy to parse visually, especially on black/white displays. 

That being said,

- I agree that ``self.subreddit_name`` is probably better than ``selfpost`` for the reason that you gave. Also IMO ``self.subreddit_name`` looks bit nicer. Either way it's just a placeholder for an empty link. Like you said, this would be easy to change.
- I think the ``/u/`` adds unnecessary clutter.
- I can take or leave the second black dot. I left it out because I felt it was unnecessary, but if you think it adds readability it might be worth adding.

If you have additional ideas and are inclined, go ahead and try them out and report back with before/after screenshots!

zambonin:
Hello, no problem with that.

I [added the `self.subreddit_name`](https://cloud.githubusercontent.com/assets/10156699/10473787/bcd37dd4-7205-11e5-95cb-0f1581f993f5.png) bit and also took the liberty of adding brackets in every flair so it's consistent. If that's not to your liking, it can be removed effortlessly.

I have also added an [exit prompt](https://cloud.githubusercontent.com/assets/10156699/10473786/bcd34b2a-7205-11e5-83fe-c8bcc7d3de14.png). Saves the hassle of logging in again if you hit `q` unintentionally.

I refrained from adding the extra black dot. It would lead to creating a `get_separator` method so [this line](https://github.com/michael-lazar/rtv/blob/master/rtv/subreddit.py#L213) could use the `config.unicode` setting...

[...]

PR #156, opened 2015-10-14
Logout and quit prompts and other small stuff.

zambonin:
According to #153.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -131,9 +131,12 @@ class BaseContent(object):
         data['flair'] = flair
         data['url_full'] = sub.url
 
+        if flair is not None and flair[0] != '[':
+            data['flair'] = '[{}]'.format(flair.strip())
+
         if data['permalink'].split('/r/')[-1] == data['url_full'].split('/r/')[-1]:
             data['url_type'] = 'selfpost'
-            data['url'] = 'selfpost'
+            data['url'] = 'self.{}'.format(data['subreddit'])
 
         elif reddit_link.match(data['url_full']):
             data['url_type'] = 'x-post'
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -140,7 +140,8 @@ class BaseContent(object):
 
         elif reddit_link.match(data['url_full']):
             data['url_type'] = 'x-post'
-            data['url'] = 'x-post via {}'.format(data['subreddit'])
+            data['url'] = 'x-post via {}'.format(strip_subreddit_url(
+                                                 data['url_full'])[3:])
 
         else:
             data['url_type'] = 'external'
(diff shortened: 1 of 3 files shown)


rtv/content.py
```diff
@@ -131,9 +131,12 @@ def strip_praw_submission(sub):
         data['flair'] = flair
         data['url_full'] = sub.url
 
+        if flair is not None and flair[0] != '[':
```
michael-lazar: This will crash if the flair is an empty string, I would suggest using flair.startswith() instead.

Do some of the flairs already have brackets added? I haven't seen any.

rtv/content.py
```diff
@@ -131,9 +131,12 @@ def strip_praw_submission(sub):
...
+            data['flair'] = '[{}]'.format(flair.strip())
+
         if data['permalink'].split('/r/')[-1] == data['url_full'].split('/r/')[-1]:
             data['url_type'] = 'selfpost'
-            data['url'] = 'selfpost'
+            data['url'] = 'self.{}'.format(data['subreddit'])
```
michael-lazar: Might as well do the same thing for x-posts to maintain a consistent style.

rtv/oauth.py
```diff
@@ -106,7 +106,7 @@ def authorize(self):
         except (praw.errors.OAuthAppRequired, praw.errors.OAuthInvalidToken):
             show_notification(self.stdscr, ['Invalid OAuth data'])
         else:
-            message = ['Welcome {}!'.format(self.reddit.user.name)]
+            message = ['Welcome, {}!'.format(self.reddit.user.name)]
```
michael-lazar: Unnecessary change

rtv/content.py
zambonin: All flairs from /r/bestof have tags. Will fix.

rtv/content.py
zambonin: If I understood correctly what you meant, the last change addresses this.

zambonin:
Should be all fixed. :)

michael-lazar:
Merged, thanks again for the PR!

shaggytwodope:
@michael-lazar @zambonin I understand this pr and such been merged already. But would it be worth considering a config option to bypass this behavior? As I rather enjoyed the previous behavior.

michael-lazar:
@shaggytwodope. I do like the idea of having a confirmation dialog. Skipping the *Enter* and quitting right after `qy` would be an easy way to bring the keystrokes down from 3 to 2. Or perhaps you could use capital `Q` to force quit without the confirmation. Do you think either of these options would be sufficient?

shaggytwodope:
@michael-lazar  Oh indeed a capital q would would be the best of both worlds I think.

zambonin:
@shaggytwodope @michael-lazar It's implemented. I'm currently trying to solve another issue and will make the PR soon.

PR #158, opened 2015-10-15
New screenshot and code width

zambonin:
Few aesthetic changes so the code is completely linted. 

Also updated the screenshot to match the new features -- specially the post count -- and the pypi badge (it works under 3.5 without a hitch).


michael-lazar:
Looks good :+1: I appreciate that you kept the style of the screenshot as close to the original as possible. I generally don't encourage PEP8 changes but these ones look sane and don't hurt the readability.

PR #167, opened 2015-10-26
After update to version 1.6.1 Exception UnicodeEncodeError closes the application

marc-queiroz:
Any SUBREDDIT with special characters like u'\xe3' are triggering an error at line 141, data['flair'] = '[{}]'.format(data['flair'].strip()) .
The solution was to add u'[{}]' to the concatenation.
The original error message: UnicodeEncodeError: 'ascii' codec can't encode character u'\xe3' in position 7: ordinal not in range(128) .


michael-lazar:
Thanks, looks good!

Issue #64, opened 2015-04-02
Rebindable Keys


Ram-Z:
Allow rebinding keys in the config file.

tobywhughes:
I'll take a look at this today and see how hard it would be.

michael-lazar:
Here's an example of another Python project that handles this pretty well. https://github.com/bpython/bpython/blob/master/bpython/sample-config#L56

rpesche:
Hello !
I wan't to work on this enhancement.
Is there any work done about it ?

michael-lazar:
Hi rpesche! Thanks for offering to help, there has been no work done on this so far AFAIK.

I'm currently in the middle of refactoring a bunch of large components, including how the config is loaded and stored internally. The branch is here https://github.com/michael-lazar/rtv/blob/future_work/rtv/config.py. Because things will be changing within the next few weeks, I would recommend waiting until the next update before starting to implement anything new on top of the master branch.

rpesche:
Ok, I'm waiting for the merge of your branch !

michael-lazar:
Hey, FYI I just merged everything into master. I'll be doing a lot of testing over the next couple of days, but will only be pushing bugfixes. You're good to go now if you want to start working on this.

rpesche:
Ok ! I'm working on it.
I had made the branch rpesche/rtv to store my work

michael-lazar:
10 months and 3 failed attempts later, it's finally in! #195

PR #182, opened 2016-01-02
Custom keybinding

rpesche:
#64
This add possibility of user key bindings.

It works but may need some improvment :
- [x] Pass test
- [ ] Documentation
- [ ] Better binding's name
- [ ] merge some bindings (Same key to refresh submission, subreddit and subscription)


michael-lazar:
Hi! Thanks for the PR, I've got a couple of points that I think are important to hit before this gets merged. In addition to the points you already listed..

- There should be some new unit tests to cover this
- The default bindings should be defined in **rtv.cfg** instead inside the code. This follows the convention of the other configuration settings.
- All of the current keys should be accounted for. E.g. you can move the cursor with the arrow keys but those currently aren't listed in the default bindings.
- Instead of defining the bindings twice (once in Controller.register() and once here, we should have a scheme where they're only defined once. Maybe come up with aliases and register those instead. 
E.g.
```python
@SubredditController.register(KeyMap.PROMPT_SUBREDDIT)
def prompt_subreddit(self):
    ...
```
It has the added benefit of letting us merge bindings (you could use **KeyMap.REFRESH** for multiple functions). I'm unsure exactly how this would be implemented though.

michael-lazar:
Hey I took some ideas from this and implemented an alternate PR #195 that met the rest of the requirements.  Sorry this didn't get merged, but it was super helpful as a reference and helped me solidify what needed to get done.

rpesche:
Ok ! Nice !

The truth is that I was slowly forgetting this pull request.
Thanks to make it real !

PR #185, opened 2016-01-07
Add coloured tag for sticky comments.

### LINKED ISSUE #184: [Feature request] Don't display sticky posts the same as regular posts
It would be great if sticky posts would be displayed differently from regular posts when the page is sorted by "hot".

So implementing one/both of these features:
* Stickied posts are in another colour
* Stickied posts are not numbered. Numbering starts on first "regular" post.

Don't know if praw supports checking if posts are stickied but if it does this might not be a hard problem to solve. I might even try to solve it myself.

5225225:
Fixes #184.

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -153,6 +153,7 @@ class Content(object):
         data['likes'] = sub.likes
         data['gold'] = sub.gilded > 0
         data['nsfw'] = sub.over_18
+        data['stickied'] = sub.stickied
         data['index'] = None  # This is filled in later by the method caller
 
         if sub.url.split('/r/')[-1] == sub.permalink.split('/r/')[-1]:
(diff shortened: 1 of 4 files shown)


michael-lazar:
The code looks good, but it doesn't look like it's working for me

image

Any ideas?

5225225:
This is for sticky comments (feature that was added quite recently).

screenshot

2016-01-09-084221_1920x1080_scrot is what stickied posts look like. I just pushed that to my repo.

By the way, some files seem to not have a newline at the end of them, which means whenever I write a file that I have edited, it also touches the bottom of the file to add the newline.

michael-lazar:
Oh neat, I didn't know there were comment stickies! 

I have automatic newlines turned off through pycharm, but it doesn't matter to me one way or the other.

Issue #180, opened 2015-12-30
[Feature request] Support for saving posts/comments, viewing saved posts/comments


daturkel:
Reddit supports saving posts to a single list tied to the user, for instance, [my saved posts](https://reddit.com/user/daturkel/saved). It would be nice if RTV allowed for the saving of posts (shortcut from sub view or from post comment view) and a way to view those saved items in a list (like the /user/[username]/saved page on reddit).

Potential obstacle: Reddit allows for the saving of both posts and comments.
a) Since comments are saveable, it might be necessary for two separate shortcuts in the comment view—one to save the highlighted comment and one to save the entire post.
b) Viewing the list of saved items might need to be formatted in a way that's slightly different from a typical subreddit view since it can include both posts and comments.

michael-lazar:
Thanks for bringing this up, I agree that it would be a nice feature to add.

PRAW has a function **r.user.get_saved()** that we could use to retrieve the user's saved items.

On point A, I think that we can use the same logic as when you press "o" in the *comment view*. If no comment is selected i.e. the submission is selected, you save the submission. If a comment is selected, you save the comment. In both cases we would need to add a flair, like the NSFW flair, to indicate that the item has been saved.

On point B, I think the most straightforward thing would be to add support for comments to the *subreddit view* and put in a new alias (e.g. **/r/saved**) similar to how **/r/me** works to view your submissions.

daturkel:
Agreed with both of your solutions! Looking forward to seeing this somewhere down the line.

alexk307:
I've got point B, working on Point A!

alexk307:
@michael-lazar which key would be the best to bind to this saving functionality?

alexk307:
I've also left a few questions in the PR about how to handle a few specifics. Thanks!

michael-lazar:
Hi Alex! You should sync up with @tyjak, he made a pull request for the same issue a few days ago https://github.com/michael-lazar/rtv/pull/187. Maybe you guys can pool your PR's together and come up with the best solution? To answer your questions:

- I like tyjak's keybinding of <kbd>w</kbd>. I have no strong inclination either way, but I like w because it's easy to type and it's close to the other modifier keys (upvote, downvote, edit, delete, etc.)
[...]

PR #187, opened 2016-01-17
Implement save/unsave post + saved list display

tyjak:
Hi, I needed the save functionallity so I added 'w' to save/unsave a post and /r/saved to show the saved list.

PR #188, opened 2016-01-20
Feature: View and save posts/comments #180

alexk307:
Code from issue #180.

This PR allows for users to type `/saved` in the subreddit prompt and view their saved posts. This DOES NOT address saved comments due to the nature of displaying comments as posts. I will add that feature next with some guidance from @michael-lazar.

You can also type `y` on a post or a comment to save it to your authenticated user. It will also add a label to the post. Currently the label will be applied on the next load. I'm not sure the best way to handle this? Should it refresh immediately or wait for the user to do a refresh?

Please let me know if there are any issues, thanks!


alexk307:
Closing because #187

PR #201, opened 2016-03-24
Made it clear you should install rtv as root

nloomans:


tobiasvl:
Why should you install rtv as root? That depends entirely on the environment, doesn't it?

nloomans:
It is required to install as root on Linux, not sure about OS X

michael-lazar:
So I used to explicitly say to use ``sudo pip install rtv`` in the README. A lot of people from the python community complained that it was promoting bad practice. I don't necessarily agree with that, but  the truth is that there's no reason that you *need* to install as root. It's just the easiest way to do it.

Thank you for the PR though @nloomans, I appreciate your involvement!

PR #223, opened 2016-05-25
Automatic viewer

Ziul:
Whatever a link with a image (jpg, png, gif, jpeg, bmp) is clicked it will be checked if there is a local program configured to open the image instead the browser. If not defined or there isn't any, calls browser as default. Tests included, but Travis don't have a default program to `*.png` ¬¬.

**Notes:**
- Some PEP8 fixes into the original code.
- Tests updated to send `config` instead of `ascii`only.

### DIFF
--- a/rtv/terminal.py
+++ b/rtv/terminal.py
@@ -37,10 +37,11 @@ class Terminal(object):
     RETURN = 10
     SPACE = 32
 
-    def __init__(self, stdscr, ascii=False):
+    def __init__(self, stdscr, config):
 
         self.stdscr = stdscr
-        self.ascii = ascii
+        self.config = config
+        self.ascii = config['ascii']
         self.loader = LoadScreen(self)
         self._display = None
 
@@ -271,7 +272,7 @@ class Terminal(object):
         # Cut off the lines of the message that don't fit on the screen
         box_width = min(box_width, n_cols)
         box_height = min(box_height, n_rows)
-        message = message[:box_height-2]
+        message = message[:box_height - 2]
 
         s_row = (n_rows - box_height) // 2
         s_col = (n_cols - box_width) // 2
@@ -299,6 +300,21 @@ class Terminal(object):
 
         return ch
 
+    def open_image(self, url):
+        """
+        Open the given image url with default image viewer
+        """
+
+        if self.display:
+
+            command = self.config['image_view'].split() + [url]
+            with self.loader('Opening page in a new window'), \
+                    open(os.devnull, 'ab+', 0) as null:
+                p = subprocess.call(command, stdout=null, stderr=null)
+        else:
+            return False
+        return True
+
     def open_browser(self, url):
         """
         Open the given url using the default webbrowser. The preferred browser
@@ -324,6 +340,15 @@ class Terminal(object):
         but are not detected here.
         """
 
+        images_pattner = ['.jpg', '.png', 'gif', 'jpeg', 'bmp']
+        if any(patt in str(url) for patt in images_pattner):
+            try:
+                if self.open_image(url):
+                    return
+            except:
+                # FAiled to open local. Keep opening with browser
+                pass
+
         if self.display:
             command = "import webbrowser; webbrowser.open_new_tab('%s')" % url
             args = [sys.executable, '-c', command]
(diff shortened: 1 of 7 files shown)


rtv/terminal.py
```diff
@@ -6,10 +6,16 @@
...
 import curses
+import mailcap
+import mimetypes
+import itertools
 import webbrowser
 import subprocess
 import curses.ascii
+from re import search
 from curses import textpad
+from xdg import BaseDirectory
```
michael-lazar: These import statements should be moved to the next section with the other third party imports.

rtv/terminal.py
```diff
@@ -26,6 +32,9 @@
     from six.moves import html_parser
     unescape = html_parser.HTMLParser().unescape
 
+# programs that don't support open file by url
+black_list_programs = ['gpicview.desktop', 'pinta.desktop']
```
michael-lazar: Where did this list come from. I'm worried that if these are just the ones that you've found so far, there could be dozens of other programs that need to be blacklisted.

rtv/terminal.py
```diff
@@ -145,6 +155,40 @@ def suspend():
...
+    @staticmethod
+    def select_program(file_url):
+        """
+        If there is a valid program to open the link, will return it, otherwise
+        returns None and the link will be opened with browser.
+        """
+        mimetype = mimetypes.guess_type(file_url)[0]
+        d = mailcap.getcaps()
+        try:
+            m = mailcap.findmatch(d, mimetype)
```
michael-lazar: ``m`` is not very descriptive, would be more clear if it was something like ``cmd, entry = mailcap.findmatch(d, mimetype)``.

rtv/terminal.py
```diff
@@ -145,6 +155,40 @@ def suspend():
...
+        except AttributeError:
+            return None
+        candidates = []
+        if m[0] == None:
```
michael-lazar: In general I'm not quite following this logic.

- convention is to use ``m[0] is None`` instead of ``m[0] == None``
- If m[0] is not None, i.e. if mailcap finds a match, candidates is an empty list and the function returns None?
- Why search for matches manually, how is that different than what mailcap.findmatch() is already doing? 
- Adding d[key][0].items() and d[key][0].keys() looks like a typo. 

For example,
  ```
  >>> d['image/png'][0]
  {'description': 'PNG Image',
  'nametemplate': '%s.png',
  'test': 'test -n "$DISPLAY"',
  'view': "feh '%s'"}
  ```
  After you append to the list, you get
  ```
  [('test', 'test -n "$DISPLAY"'),
   ('nametemplate', '%s.png'),
   ('description', 'PNG Image'),
   ('view', "feh '%s'"),
   'test',
   'nametemplate',
   'description',
   'view']
  ```
This makes candidates look like a jumbled mess of strings when you do ``list(itertools.chain(*candidates))``
- What is the purpose of using the xdg library at all? I'm not at all familiar with how it works, but it's not clear to be what information is being gained.

rtv/terminal.py
```diff
@@ -299,6 +343,21 @@ def show_notification(self, message, timeout=None):
...
         return ch
 
+    def open_image(self, url):
+        """
+        Open the given image url with default image viewer
+        """
+
+        if self.display:
+
+            command = [Terminal.select_program(str(url))[0]] + [url]
```
michael-lazar: ``str`` is not python2/python3 compatible, why is this necessary

rtv/terminal.py
```diff
@@ -324,6 +383,14 @@ def open_browser(self, url):
         but are not detected here.
         """
 
+        if Terminal.select_program(str(url)):
+            try:
+                if self.open_image(url):
+                    return
+            except:
+                # Failed to open local. Keep opening with browser
+                pass
```
michael-lazar: Should catch specific errors and potentially write a log message

rtv/terminal.py
```diff
@@ -405,7 +473,8 @@ def open_editor(self, data=''):
             # in order to read the changes made by some editors (gedit). w+
             # mode does not work!
             with codecs.open(fp.name, 'r', 'utf-8') as fp2:
-                text = ''.join(line for line in fp2 if not line.startswith('#'))
+                text = ''.join(
```
michael-lazar: If you look in the *.pylintrc* file, I set the line limit to 100 in order to be a bit more lenient in cases like this where it breaks 79 by a character or two. IMO it's cleaner + more readable to keep it on a single line

rtv/terminal.py
```diff
@@ -324,6 +383,14 @@ def open_browser(self, url):
         but are not detected here.
         """
 
+        if Terminal.select_program(str(url)):
```
michael-lazar: With static methods, you can still do ``self.select_program(str(url))``

michael-lazar:
I picked this up and am working on it here https://github.com/michael-lazar/rtv/tree/custom_commands2

PR #226, opened 2016-05-26
Add "links from:" context menu for top posts

apex-meme-lord:


michael-lazar:
Hi, thanks for the PR. I appreciate how clean and straightforward your code changes are!

The one comment that I have is that I don't like using the <kbd>x</kbd> key for this. At this point in the project I would like to reserve the remaining keys for large functionality additions. Here are some alternate ideas:

- Double-pressing 2, i.e. pressing <kbd>2</kbd> when the **[2]top** is already highlighted, opens up the selection menu.
- Add a new item to the header menu, **[6]more** or **[6]other**, that lists all of your additional options when you press it.

Also, as long as were doing this it would be good to extend to the *controversial* sort order too.

apex-meme-lord:
@michael-lazar: Really great input.  I'll tinker around with the alternatives and re-open this when I have some time.

Issue #78, opened 2015-04-05
assign custom commands to run when opening images/gifs/videos


oatberry:
I'd love it if I could define custom commands to run when, for example, opening a link that ends with .gif: `mpv %u` or something like that. Or `mpv ytdl://%u` for youtube links.  Also `curl %u | feh -` for images.

michael-lazar:
This is a cool idea, I could see defining a new section in the config file. Something like...

```ini
[browsers]
*.gif="mpv %u"
youtube="mpv ytdl://%u"
imgur="curl %u | feh -"
```

This would also be a good way to tie in @McHearty's idea #70 of using w3m.

McHearty:
@michael-lazar I really like this idea, good resolution for #70

obosob:
There is a pretty simple workaround for this.

    BROWSER=urlscript rtv

and in urlscript:

    #!/bin/bash
    case $1 in
        http*://*youtube.com/watch?*|http*://youtu.be/watch?*)
             mpv "ytdl://$1"
        ;;
        *.gif|*.gifv)
             mpv "$1"
        ;;
        *.jpg*|*.png*)
            feh -. "$1"
        ;;
    esac

michael-lazar:
This is very nice! I've been slowly working on building this into the python source using a config file. It's mostly finished, but I have a few roadblocks with imgur:

1. ``.gifv`` is a custom imgur extension that's not recognized by any command-line applications. This one's not so bad because you can replace the ``.gifv`` with ``.webm`` and the link will still open.
2. Half of the imgur links look like this ``http://imgur.com/pCmsLnV`` and don't link directly to the image. I've been trying to guess at file extensions but I need a way to get the actual file type.
3. I haven't found a good way to open imgur albums, which represent a significant number of posts.

Any ideas?

shaggytwodope:
@michael-lazar Not sure how it would work with links, but there is https://github.com/ahupp/python-magic which can find out the file type. MIGHT be of some use to reference at least. There is also https://docs.python.org/2/library/imghdr.html which does images. But not sure how this would relate to links... unless it's cached somehow. 

Just tossing ideas, punch me later lol.

obosob:
@michael-lazar just had a quick poke around the source, and there is an element in the <head> as follows:

    <link rel="image_src" href="http://i.imgur.com/pCmsLnV.jpg"/>

Which is a link to the image itself, albums aside for now.

so you could do a quick hack in the script:

[...]

PR #228, opened 2016-06-13
Add an optional 'handlers' section to the config file for special han…

llvtt:
…dling of certain mimetypes.

I can clean this up any way you prefer; I just wanted to gauge interest in this feature first.

**Inspiration**
iTerm (terminal emulator I use on my mac) recently gained the ability to display images directly in the terminal: http://iterm2.com/documentation-images.html. I thought, why not make this available in RTV?

**Configuration**
Something like this goes into your rtv.cfg:

    [handlers]
    ; Special handler to open images.
    ; <handler name> = <list-of-mimetypes>
    rtv.mime_handlers.iterm2_image_viewer = image/jpeg,image/jpg,image/png,image/gif

**Results**

bildschirmfoto 2016-06-12 um 16 59 03
bildschirmfoto 2016-06-12 um 16 58 59

This isn't just about images, though. This allows users to handle different mimetypes with python functions from RTV. I could also imagine a function to handle audio without leaving the terminal, for example. Also, as long as the function is importable, rtv can use it. It doesn't have to be defined in rtv's own source code.

Hope someone else finds this hack interesting/useful. Let me know if you think something needs to be cleaned up/changed.

### DIFF
--- a/rtv/__main__.py
+++ b/rtv/__main__.py
@@ -40,12 +40,13 @@ def main():
     sys.stdout.write('\x1b]2;{0}\x07'.format(title))
 
     args = Config.get_args()
-    fargs, bindings = Config.get_file(args.get('config'))
+    fargs, bindings, handlers = Config.get_file(args.get('config'))
 
     # Apply the file config first, then overwrite with any command line args
     config = Config()
     config.update(**fargs)
     config.update(**args)
+    config.set_mime_handlers(handlers)
 
     # If key bindings are supplied in the config file, overwrite the defaults
     if bindings:
@@ -88,7 +89,7 @@ def main():
             if not config['monochrome']:
                 Color.init()
 
-            term = Terminal(stdscr, config['ascii'])
+            term = Terminal(stdscr, config)
             with term.loader('Initializing', catch_exception=False):
                 reddit = praw.Reddit(user_agent=user_agent,
                                      decode_html_entities=False,
--- a/rtv/config.py
+++ b/rtv/config.py
@@ -123,6 +123,7 @@ class Config(object):
         default, bindings, handlers = self.get_file(DEFAULT_CONFIG)
         self.default = default
         self.keymap = KeyMap(bindings)
+        self.set_mime_handlers(handlers)
 
         # `refresh_token` and `history` are saved/loaded at separate locations,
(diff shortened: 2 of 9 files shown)


llvtt:
I made another commit to fix the existing tests, though I haven't added any for my own code. I'd first like to know if something like this is of interest to the project, and if so, I'd be happy to write tests for it. Just let me know.

llvtt:
D'oh! I just made something I thought was cool and dropped a PR here without looking at the larger context of the project. I see now that this PR more or less duplicates the effort of #223, which is using mailcap. That's a better idea than using the config file like I'm doing here, since it doesn't require writing any code on the user's part and just does the Right Thing if the user has one set up already.

@michael-lazar @Ziul, let me know if either of you need any help working on this. I recently started using RTV and am hooked, and I'd love to contribute something back.

Thanks for the great work!

michael-lazar:
Hi @llvtt, thanks for the PR! This is probably the most requested feature for rtv and there have been at least a half dozen attempts (mostly by me) so far.

I wish I could try it out but unfortunately I don't have access to a mac :cry: The iTerm viewer looks super interesting though, reminds me of something that I saw in [ranger](https://github.com/ranger/ranger/blob/master/ranger/ext/img_display.py#L181). I also like the way that you set up the config file with importable python handlers. If you don't mind, I'd like to re-open this and at least keep it around as a reference that we can pull ideas from.

llvtt:
I definitely don't mind reopening this :).

Using python handlers as I've done here is nice because we can integrate easily with the Terminal object for displaying notifications, among other uses.

However, as I mentioned before, using mailcap is great because the same mailcap file can be used across applications. Using external programs to deal with various mimetypes is also less maintenance work for this project, rather needing to add new python handler functions for each mimetype.

I'd be willing to make this PR use mailcap, but I don't want to step on anyone's toes who's already working on this.

michael-lazar:
Linking to issue https://github.com/michael-lazar/rtv/issues/78 so it shows up on the tracker.

michael-lazar:
The mailcap approach has now been released, so I think it's a good time to close this. Please let me know if you have any suggestions for improving the default mailcap settings for iTerm!

Issue #203, opened 2016-03-25
[Feature Request] local subreddit list when not logged


marteaurouge:
When hitting s , a subreddit list is popping, but not if you're not logged.

What about a config file to store locally a list of your fav subreddit, so the s feature would be useful even without using an account.

spcmd:
Yes, It would be great to have them stored locally in a plain text file. I think it would be faster to open the list too.

michael-lazar:
Cleaning out duplicate issues, discussion can continue on #322

PR #233, opened 2016-06-27
Feature: view user pages and domain listings

woorst:
Added the ability to browse pages like `www.reddit.com/u/spez/submitted` and `www.reddit.com/domain/imgur.com`.

Also, content can be sorted by time period like all, past year, past month, etc.


michael-lazar:
Hey thanks for this! I'll be testing over the next few days.

If we do this I think we should take the opportunity to get rid of the ``/r/`` at the beginning of the prompt, it looks weird typing ``/r/u/spez/submitted``. I also think it would be helpful to list out examples for all of the possible combinations, e.g. 
- ``/r/python/new``
- ``/user/spez/hot``
- ``/domain/imgur.com``
- etc

woorst:
I think things are set enough for you start taking a look at.

Yes, I agree that the prompt should be edited to reflect the new additions.

So I added code to allow sorting by time period (month, year, etc..) but right now there aren't any keyboard shortcuts to expose that functionality to the user. Wasn't sure how you'd want to go about that.

michael-lazar:
Here's a related pull request where we talked about the sorting problem https://github.com/michael-lazar/rtv/pull/226

CONTROLS.rst
```diff
@@ -44,11 +45,15 @@ In subreddit mode you can browse through the top submissions on either the front
 
 The ``/`` prompt accepts subreddits in the following formats
```
michael-lazar: I really like how you set up the prompts, looks perfect!

CONTROLS.rst
```diff
@@ -44,11 +45,15 @@ In subreddit mode you can browse through the top submissions on either the front
...
+* ``python``
 * ``/r/python``
 * ``/r/python/new``
 * ``/r/python+linux`` supports multireddits
 * ``/r/front`` will redirect to the front page
-* ``/r/me`` will display your submissions
+* ``/u/me`` will display your submissions
+* ``/u/spez`` will display submissions from any other user
+* ``/u/multi-mod/m/android`` display a multireddit curated by a user
+* ``/domain/python.org`` dispaly submissions to stated domain
```
michael-lazar: Typo in ``dispaly``

rtv/content.py
```diff
@@ -510,3 +574,52 @@ def get(self, index, n_cols=70):
...
+        self._multireddit_data = []
+
+        try:
+            self.get(0)
+        except IndexError:
+            raise exceptions.SubscriptionError('No multireddits')
+
+    @classmethod
+    def from_user(cls, reddit, multireddits, loader):
+        multireddits = (m for m in multireddits)
```
michael-lazar: Is there a reason that you didn't follow the SubscriptionContent() convention and do something like this here?

```python
multireddits = reddit.get_my_multireddits(limit=None)
```

rtv/multireddits.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
```
michael-lazar: It looks like almost this whole file is copied from **submissions.py**. If possible, I think it would be better to just add a flag to the SubscriptionPage.

```python
page = SubscriptionPage(reddit, term, config, oauth)
```
vs
```python
page = SubscriptionPage(reddit, term, config, oauth, multireddit=True)
```

I don't see a problem with using the same keybindings (``SUBSCRIPTION_SELECT``) and (``SUBSCRIPTION_EXIT``) for both because the pages are functionally the same.

CONTROLS.rst
```diff
@@ -30,6 +30,7 @@ Once you are logged in your username will appear in the top-right corner of the
 :``d``: Delete an existing post or comment
 :``i``: Display new messages prompt
 :``s``: View a list of subscribed subreddits
+:``S``: View a list of subscribed multireddits
```
michael-lazar: Should also be added to **docs.py** so it shows up when you type <kdb>?</kbd>

tests/test_content.py
```diff
@@ -240,6 +240,21 @@ def test_content_subreddit_from_name(reddit, terminal):
     assert content.name == '/r/python'
     assert content.order is None
 
+    name = '/domain/python.org'
```
michael-lazar: Since this test requires HTTP requests to run, you will need to generate a new cassette in order to get it to pass. Check out the section in the README faq about running tests and let me know if you have any questions.

rtv/content.py
```diff
@@ -367,52 +385,96 @@ def __init__(self, name, submissions, loader, order=None):
...
 
     @classmethod
-    def from_name(cls, reddit, name, loader, order=None, query=None):
+    def from_name(cls, reddit, name, loader, order=None, query=None,
+                  listing='r', period=None):
+
+        # Strip leading, trailing, and redundant backslashes
+        n = ''
+        n = ''.join([n + ''.join(list(g)) if k != '/' else '/' \
+                                       for k, g in groupby(name)]).strip(' /')
```
michael-lazar: Not sure what this is doing but it looks like it could be simplified, or at least expanded so it's easier to read. What is ``groupby()`` doing?

rtv/content.py
woorst: Two things here:
- the praw methods `get_multireddits` and `get_my_multireddits` return lists rather than generators. I turn it into a generator to maintain some similarity in the code.

- I want this method to potentially load not just the logged in user's multireddit but also any desired user. So either list of multireddits or the username should be passed to this method.

rtv/multireddits.py
woorst: True, these files are nearly identical. I was thinking perhaps have a separate file for different kinds of reddit pages might be easier to manage. Your alternative works too.

How would it work to use the same keybindings (`SUBSCRIPTION_SELECT`) to select two different pages? Do you mean have all subscribed subreddits and multireddits be listed together on one page?

rtv/content.py
woorst: `groupby()` is used here to group redundant `/`'s and flatten them. This is to catch those edge cases in the unit test. Also, typos like `/r/python//` or `/r/python//top` do what you'd want them to do.

Yes, this segment can be written more clearly.

michael-lazar: Oh I see. I've never seen anybody use ``groupby`` like that before!

Here's an alternative method that may be clearer
```
name = '/'.join(seg for seg in name.split('/) if seg)
```

woorst:
@michael-lazar Hey, can you take a look at this now.

michael-lazar:
Hey, thanks for addressing my comments. I went ahead and made a couple more changes and then merged this in. Overall I didn't change any of the functionality, just a couple of code style things.

- Reverted the name change from "ListReddit" back to "Subscription". The change was causing a bunch of unnecessary diffs and I'm not sure that it was adding much clarity. I think at this point it's better to keep the legacy name for compatibility.
- Moved the "My Subscriptions" / "My Multireddits" logic away from the lambda function definitions and into SubscriptionContent. The reasoning is that the classes in content.py should be responsible for defining the interface with praw.
- Added a bunch more test cases for the subreddit prompt.
- Exposed the ability to sort by day/week/hour/etc by adding a dash to the end of the url.

/r/**{subreddit}**/**{order}**-**{period}**
/r/python/controversial-day

woorst:
Thanks for the merge.

As to why I had the SubscriptionContent receive a function as an input: there are other instances when you might like to display a list of reddits, e.g. `reddit.get_{new_subreddits,subreddit_recommendations}` or as requested in #203. It might be simpler to just pass a function that generates the list than to try and anticipate all possible situations.

PR #234, opened 2016-06-27
refactored for better readability

nagracks:
Tried on two files *rtv/config.py* and *rtv/content.py*


michael-lazar:
Hi, thanks for the PR. This project follows the PEP 8 style guide (more or less), which guides *against* using extra spaces to align statements like this.

Reference: https://www.python.org/dev/peps/pep-0008/#pet-peeves

PR #237, opened 2016-06-30
Set terminal title to reflect current page

woorst:


michael-lazar:
Looks good for subreddits, but I think it gets too cluttered when it prints the entire submission URL.

image
image

How do you feel about just doing this for subreddits, unless you have any other ideas? Also I would suggest putting the **rtv vX.Y.Z** first so it doesn't get cut off if the window is too small.

woorst:
How about I ellipsize the middle of the title string if it is over so many characters?

It seems to be common append program name after the filename in most applications

michael-lazar:
Do you have any examples? I don't see many applications that edit the terminal title in the first place. The one that I did find was ``elinks``,  which I believe just uses the \<title\> element instead of the whole url.

image

woorst:
[vim](https://i.imgur.com/VchbV.png)
[ncmpcpp](http://rybczak.net/ncmpcpp/screens/screen03.png)
[mps-youtube](https://github.com/mps-youtube/mps-youtube/blob/9152509e820c2d822f5de643b264d73e07b2b7a3/mps_youtube/util.py#L122)

michael-lazar:
Fair enough, thanks for providing examples. I think that cutting to 50 characters is good enough to address my concern.

Issue #100, opened 2015-05-05
[Feature Request] Enter comments to directly open links in browser


TLATER:
Pretty much just as the title says, many comments also contain links and being unable to open those without looking for them in the comment section forces mouse usage.

michael-lazar:
I agree that this would be a great thing to have. There are two parts to the problem.

1. We need to either find a builtin library function or build some custom regular expression to search for valid links in the comment bodies. I don't anticipate this being too difficult.
2. We need a way to handle comments that containmore than one link. How would you select which link to open? I would be hesitant to add new key bindings for this unless they were very intuitive.

TLATER:
I'd suggest another mode that is accessible through submission mode. Accessing it would use right arrow, which folding currently uses. I guess folding could instead use enter, that appears more intuitive than right arrow to me (or vice-versa, but that would be inconsistent with the current keybindings).

In comment mode up and down select individual links, enter opens them, and left moves back to submission mode. There could even be more features, like instantly replying to the current comment, or showing the context.

If I dream on this might even be useful for implementing an inbox feature somewhere down the line.

mrjovanovic:
Same should be valid for selfposts, which also often have URLs.

For the regex, [here](https://mathiasbynens.be/demo/url-regex) is an overview of different options. It seems that this one is probably the best for our purposes, as it fails on mistyped URLs:

[...]

PR #240, opened 2016-07-05
Added urlview keybinding

msmith491:
1.  A shortcut of 'b' was used as the default binding for passing
comment body text to urlview (this is similar to the default of C-b in
the 'mutt' text email client)
2.  The `comment_urlview` SubmissionController function was added to
recieve the SUBMISSION_OPEN_IN_URLVIEWER keypress request.
3.  The `open_urlview` terminal function was added to handle the urlview
request.  It passes the comment body data to urlview via a Popen
process.
4.  A test case was added to ensure this new code path is executed
5.  Small formatting changes, mostly line length

This patch was designed and tested with urlview, but should work with any url extracting binary that accepts text from stdin.  I realized I missed the functionality from the mutt email client where you could easily extract a list of URLs from the email body text with a shortcut

An example of a comment where this would be handy:
https://www.reddit.com/r/news/comments/4ra3wr/christian_mingle_must_let_lgbt_singles_use_dating/d4zgobc
But pretty much any comment with many links would fit the need.

Will likely require updating the README as well to instruct users to install urlview (or an equivalent) in order to use this feature.

Please let me know if there are any modifications necessary to be more inline with the repository standards :)


msmith491:
Additionally it might be handy if rtv respected the contents of a ~/.mailcap file to deal with links of various types.  For example the user might want to open a `text/html` link in `w3m`, but `image/*` links they might want opened with an image-viewer like `feh`.

If there's interest I could make a more general pull request with those features.

michael-lazar:
Very cool! I think this is a perfect way to solve issue https://github.com/michael-lazar/rtv/issues/100.

For documentation we need to...

- Add a section to the **README**, using the same format as the *Editor* and *Web Browser* sections.
- Add the new keybind to **CONTROLS.rst**
- Add the new environment variable to **scripts/rtv.1.template**

and we should be all set.

As for mailcap, a couple of different people have looked into it. See the discussion thread here https://github.com/michael-lazar/rtv/issues/78

rtv/terminal.py
```diff
@@ -438,6 +439,20 @@ def open_editor(self, data=''):
             else:
                 _logger.info('File deleted: %s', filepath)
 
+    def open_urlview(self, data):
+        urlview = os.getenv('URLVIEWER') or os.getenv('RTV_URLVIEWER') \
```
michael-lazar: Is ``$URLVIEWER`` is a globally adopted variable (like ``$BROWSER`` and ``$PAGER``)? I did a quick google search and couldn't find anything. Ideally we would be able to use something standard, but if nothing exists then I think this is ok.

msmith491: I made `$URLVIEWER` up because I couldn't find anything else :).  I'm happy enough to drop it for just `$RTV_URLVIEWER` to ensure we're not polluting the environment namespace

rtv/terminal.py
```diff
@@ -438,6 +439,20 @@ def open_editor(self, data=''):
...
+        try:
+            with self.suspend():
+                p = subprocess.Popen([urlview],
+                                     stdin=subprocess.PIPE)
+                try:
+                    p.communicate(input=data)
+                except KeyboardInterrupt:
+                    p.terminate()
+        except OSError:
+            self.show_notification('Could not open urls with urlview')
```
michael-lazar: The notification should show the command defined by the environment variable if applicable

michael-lazar:
Crashes for me on python 3.4

```
Traceback (most recent call last):
  File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "rtv/__main__.py", line 137, in <module>
    sys.exit(main())
  File "rtv/__main__.py", line 118, in main
    page.loop()
  File "rtv/page.py", line 70, in loop
    self.controller.trigger(ch)
  File "rtv/objects.py", line 581, in trigger
    return func(self.instance, *args, **kwargs) if func else None
  File "rtv/subreddit.py", line 91, in open_submission
    page.loop()
  File "rtv/page.py", line 70, in loop
    self.controller.trigger(ch)
  File "rtv/objects.py", line 581, in trigger
    return func(self.instance, *args, **kwargs) if func else None
  File "rtv/submission.py", line 155, in comment_urlview
    self.term.open_urlview(comment)
  File "rtv/terminal.py", line 450, in open_urlview
    p.communicate(input=data)
  File "/usr/lib/python3.4/subprocess.py", line 941, in communicate
    self.stdin.write(input)
TypeError: 'str' does not support the buffer interface
```

msmith491:
- Added the documentation sections
- Updated the notification message to include binary information
- Fixed the python3 crash

michael-lazar:
Merged! Thanks so much for the contribution!

PR #255, opened 2016-07-22
urlview command works on submission as well as comments

woorst:


rtv/submission.py
```diff
@@ -150,7 +150,7 @@ def delete_comment(self):
     @SubmissionController.register(Command('SUBMISSION_OPEN_IN_URLVIEWER'))
     def comment_urlview(self):
         data = self.content.get(self.nav.absolute_index)
-        comment = data.get('body', '')
+        comment = data.get('body') or data.get('text') or data.get('url')
```
michael-lazar: ``data[url]`` is what's printed on the screen, so it doesn't have to be an an actual url. E.g. it could be ``self.videos``. I think you want ``data[url_full]`` instead.

PR #265, opened 2016-07-28
Parse complete imgur albums

woorst:
Multipage imgur albums can be parsed now.

One thing to note: albums can contain mixed content, e.g. .gifv and .jpg. Not sure what mime type to use for that.

### DIFF
--- a/rtv/mime_parsers.py
+++ b/rtv/mime_parsers.py
@@ -144,17 +144,22 @@ class ImgurAlbumMIMEParser(BaseMIMEParser):
             </a>
         </div>
     """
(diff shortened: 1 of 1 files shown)


michael-lazar:
Hey thanks, could you post some examples of imgur links that this fixes so I can add test cases?

michael-lazar:
I was also thinking about mixed content and couldn't come up with a good solution, perhaps it would be best to open those ones with the web browser for now.

woorst:
The `/r/diy` subreddit frequently has large imgur albums, e.g. `https://imgur.com/a/q8Vv4`

michael-lazar:
Looks like it fails for single picture galleries, e.g. http://imgur.com/gallery/0YhKlLr doesn't have a corresponding album.

PR #268, opened 2016-08-02
Added a shortened URL expander

edridgedsouza:
I believe this edit will allow RTV to accept redd.it (and bit.ly and other URL shorteners)

### DIFF
--- a/rtv/subreddit.py
+++ b/rtv/subreddit.py
@@ -11,6 +11,7 @@ from .objects import Navigator, Color, Command
 from .submission import SubmissionPage
 from .subscription import SubscriptionPage
 from .exceptions import TemporaryFileError
+from urllib.request import urlopen # For parsing shortened URLs
 
 
 class SubredditController(PageController):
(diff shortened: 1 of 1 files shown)


edridgedsouza:
Hmm. Looks like this build was unsuccessful. I'm really not too familiar with the intricacies of Python, but I do believe that something in the form of the few lines I added will allow RTV to accept shortened links. You could also put the lines I added within a conditional regex checker to prevent it from expanding every single URL it gets. Again, I'm relatively new at Python so if there's a reason why this build seems to fail, please let me know.

michael-lazar:
Thanks, but I have to reject this for a couple of reasons

- The reason the build is failing is because the code block that you added isn't valid python syntax. You can only use ``break`` inside of a loop, it doesn't mean anything in the context of ``try...except...``
- ``urllib.request.urlopen`` doesn't exist in python 2, it's called ``urllib.urlopen`` instead. That's one of the reasons it's nice to use ``requests``, because it's compatible across all python versions.
- You can use HTTP **HEAD** instead of **GET** to request the page headers only and follow redirects without loading the whole page.
- I don't like the overhead of making an extra HTTP request every time you load a submission. Because the only time you can submit a shortened url is at start-up (using the **--link** flag), I think it's better for now to just check that link rather than every single submission.

Since it was only a couple of lines, I made an alternate PR that addresses these points. https://github.com/michael-lazar/rtv/pull/276

PR #275, opened 2016-08-07
Use youtube-dl to check for multimedia links

woorst:
I noticed you have `youtube-dl` in the example mailcap file. Since the program can download video from various sites it would be nice to use to check for videos.


xelxebar:
@woorst I figured out how to do this fairly generically so that things "just work" in general with `youtube-dl`. You might be interested. In my configuration, all MIME types just push through `xdg-open`. Here's the relevant slice of my `.mailcap`:

    image/*; xdg-open "%s" >/dev/null 2>&1;; test=test -n "$DISPLAY";
    video/*; xdg-open "%s" >/dev/null 2>&1;; test=test -n "$DISPLAY";
    application/*; xdg-open "%s" >/dev/null 2>&1;; test=test -n "$DISPLAY";

Some configurations have a catch-all "*/*" line, but not all programs understand this (*e.g.* mutt). I then have `xdg-open` configured to open `http` and `https` URIs with a custom script that calls into `youtube-dl` as appropriate:

    $ xdg-mime query default x-scheme-handler/http
    httpx-scheme-handler.desktop

    $ xdg-mime query default x-scheme-handler/https
    httpx-scheme-handler.desktop

The above commands just look at the `~/.local/share/applications/mimeapps.list` file. Now `.desktop` file referenced in the above commands is a custom guy I wrote:

    $ cat ~/.local/share/applications/httpx-scheme-handler.desktop
    [Desktop Entry]
    Name=HTTP(S) Viewer
    Comment=Open http(s) URIs with the appropriate application
    Type=Application
    Exec=~/.local/share/applications/scripts/httpx-scheme-handler.sh %u
    Categories=Application;Network;WebBrowser;
    MimeType=x-scheme-handler/http;x-scheme-handler/https;

I then have `http-scheme-handler.desktop` and `https-scheme-handler.desktop` symlinked to this just in case. The really relevant line here is the `Exec` one which calls into my custom script. Here it is in all its glory:

    #!/usr/bin/env bash
    
    function mime_open {
        local mime=${1}; shift
        local args=("${@}")
        local dex=""
        local exec=""
        local basename=$(\xdg-mime query default "${mime}")
        local -A dirname=(
            [user]="${HOME}/.local/share/applications"
            [system]="/usr/share/applications"
            )
        
        [[ -z "${basename}" ]] && return -1
        
        local key
        for key in user system; do
            dex=${dirname[$key]}/${basename}
            [[ -r "${dex}" ]] && break
            dex=""
        done
    
        [[ -z "${dex}" ]] && return -1
    
        local line
        while read -r line; do
            if [[ "${line}" =~ ^Exec= ]]; then
                exec=${line#Exec=}
                exec=${exec/%%?/${args[@]}}
            fi
        done <"${dex}"
    
        [[ -n "${exec}" ]] && ${exec}
    }
    
    function youtube_dl_extractable {
        local url=${1}
        \youtube-dl --quiet --no-warnings --simulate "${url}" 2>/dev/null
    }
    
    function open_www_page {
        local mime=${1}
        local url=${2}
    
        if  $(youtube_dl_extractable "${url}"); then
            \youtube-dl --quiet --output - "${url}" \
            | mime_open 'video/*' -
        else
            mime_open "${mime}" "${url}"
        fi
    }
    
    function open {
        local mime=${1}
        local url=${2}
        
        case "${mime}" in
            "text/html") open_www_page "${mime}" "${url}";;
            *)           mime_open "${mime}" "${url}";;
        esac
    }
    
    function mime {
        local url=${1}
        local content_type=""
        local mime=""
    
        local field=""
        while read -r field; do
            field=${field/%$'\r'}
            if [[ "${field}" =~ ^Content-Type: ]]; then
                content_type=${field#Content-Type: }
                mime=${content_type%%;*}
            fi
        done < <(\curl --silent --location --head "${url}")
    
        echo ${mime}
    }
    
    function main {
        local url=${1}
        local mime=$(mime "${url}")
    
        open "${mime}" "${url}"
    }
    
    main "${@}"

If the script's a litle hard to understand, all it really does is grab the MIME type from the HTTP headers and then opens the URI with the relevant XDG-configured application. In addition, if the URI has youtube-dl extractable content, then it streams the video to your XDG-configured player.

The benefit of going to all this trouble is that now you can just do `xdg-open <some-url>` and then everything just kind of works. For example, direct jpeg links end open up in `feh` for me, videos in `mpv`, and general web pages in `uzbl-browser`. This nicely separates out the "open a URI" logic so that it's fairly trivial to make other cli apps do the "right thing".

Anyway, hope this was somewhat useful.

**NB**: The `mime_open` function is sort of a hack as it manually parses `.desktop` files for the `Exec` line, so a bit of caution might be necessary.

PR #281, opened 2016-08-10
Add requirements.txt file, this closes #283

### LINKED ISSUE #283: Add requirements.txt file
Adding this file make easy to work on project. This file will include list of all required modules that we need for `rtv`. Any contributor can then simply make `virtualenv` for this and simply do `pip install -r requirements.txt`.

nagracks:


michael-lazar:
Please post an issue on the tracker that explains why you think this should be added to the project.

requirements.txt
```diff
@@ -0,0 +1,10 @@
+beautifulsoup4==4.5.1
+decorator==4.0.10
+kitchen==1.2.4
+mailcap-fix==0.1.3
+praw==3.5.0
+requests==2.11.0
+rtv==1.11.0
+six==1.10.0
+tornado==4.4.1
```
michael-lazar: You can get rid of tornado, that was recently removed as a dependency here #278. Other than that, looks good!

nagracks: tornado is removed

PR #286, opened 2016-08-12
Update README

nagracks:
Add complete commandline help under `Usage`


michael-lazar:
Closed because I'm trying to keep the README as short and concise as possible. I don't think that it's necessary to display all of the command line options.

PR #289, opened 2016-08-24
Make rtv  not dependeing on rtv

Fale:


michael-lazar:
Thanks, I didn't notice that!

PR #302, opened 2016-09-28
Setting the maximum comment column width

alx-k:
I opened a feature request issue for this but decided to implement it.

I've added a command line and config file argument for to set the maximum column width for submission comments. By default this is set to 80.

This makes reading comments much smoother when using a full screen terminal.

It's only my second contribution to someone else's code, so feedback on code quality is welcome :)

### DIFF
--- a/rtv/config.py
+++ b/rtv/config.py
@@ -67,6 +67,9 @@ def build_parser():
     parser.add_argument(
         '--enable-media', dest='enable_media', action='store_const', const=True,
         help='Open external links using programs defined in the mailcap config')
+    parser.add_argument(
+        '--max-comment-cols', dest='max_comment_cols', type=int,
+        help='Maximum comment column width')
     return parser
 
 
@@ -240,8 +243,10 @@ class Config(object):
             'enable_media': partial(config.getboolean, 'rtv'),
             'history_size': partial(config.getint, 'rtv'),
             'oauth_redirect_port': partial(config.getint, 'rtv'),
-            'oauth_scope': lambda x: rtv[x].split(',')
+            'oauth_scope': lambda x: rtv[x].split(','),
+            'max_comment_cols': partial(config.getint, 'rtv')
         }
+
         for key, func in params.items():
             if key in rtv:
                 rtv[key] = func(key)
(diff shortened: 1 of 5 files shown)


rtv/config.py
```diff
@@ -67,6 +67,9 @@ def build_parser():
     parser.add_argument(
         '--enable-media', dest='enable_media', action='store_const', const=True,
         help='Open external links using programs defined in the mailcap config')
+    parser.add_argument(
+        '--max-comment-cols', dest='max_comment_cols', type=int,
+        help='Maximum comment column width')
```
michael-lazar: I don't think that a command line flag is really necessary for this setting because it's obscure enough that most people won't be messing with it. I think that having it in the config file is good enough.

rtv/templates/rtv.cfg
```diff
@@ -57,6 +57,9 @@ oauth_redirect_port = 65000
 ; Access permissions that will be requested.
 oauth_scope = edit,history,identity,mysubreddits,privatemessages,read,report,save,submit,subscribe,vote
 
+; Maximum number of columns for a comment
+max_comment_cols = 80
```
michael-lazar: This should be grouped with the other general settings above the OAuth settings

michael-lazar: Would you be opposed to making the default 120? I think that it maintains the readability while allowing more text to fit on the screen at once.

rtv/submission_page.py
```diff
@@ -23,10 +23,15 @@ def __init__(self, reddit, term, config, oauth, url=None, submission=None):
...
         if url:
-            self.content = SubmissionContent.from_url(reddit, url, term.loader)
+            self.content = SubmissionContent.from_url(
+                reddit, url, term.loader,
+                max_comment_cols=config['max_comment_cols'])
         else:
-            self.content = SubmissionContent(submission, term.loader)
+            self.content = SubmissionContent(
+                submission, term.loader,
+                max_comment_cols=config['max_comment_cols'])
```
michael-lazar: You missed a reference to ``SubmissionContent.from_url()`` made from ``self.refresh_content()``.

alx-k:
Hi Michael,

Implemented those changes! I hope it's all good.

Thanks for `rtv` by the way, now I can reddit at work ;)

michael-lazar:
Glad you're using it, if you find any other issue feel free to post them!

Issue #308, opened 2016-10-07
[feature request] (top,controversial) sort by (hour,day,week,month,year,all)


ReshefElisha:
Would be nice to have the ability to go to sort all and controversial. Proposed mapping:

(only in `controversial` or `top` modes):
6 : hour
7 : day
8 : week
9 : month
0 : year
\- : all

ReshefElisha:
In the spirit of hacktoberfest I'll attempt an implementation.

michael-lazar:
Here's a related issue https://github.com/michael-lazar/rtv/issues/244 and PR https://github.com/michael-lazar/rtv/pull/226

ReshefElisha:
Thanks, I'll take a look.
Do you prefer the added menu style or extending the existing menu by adding more options?

michael-lazar:
I think that either of the options that I outlined in #266 would be good. I lean towards adding the additional sub-menu but that I can see that being more difficult to implement. I'm also open to other suggestions if you have any good ideas.

I don't think that we should just extend the menu bar horizontally like this 
```
[1]hot [2]top [3]rising [4]new [5]controversial [6]hour [7]day [8]week [9]month [0]year [-]all
```
because it becomes too cramped.

michael-lazar:
The sub-menu feature has been merged in and is accessible by double-pressing "2" or "5".

PR #312, opened 2016-10-09
Added 'links from' menu

ReshefElisha:
in reference to issue #308 

Causes odd behavior in `/r/front` subreddit so disabled in that context.

### DIFF
--- a/rtv/docs.py
+++ b/rtv/docs.py
@@ -145,3 +145,13 @@ OAUTH_SUCCESS = """\
        <p><span style="font-weight: bold">Reddit Terminal Viewer</span>
        will now log in, you can close this window.</p>
 """
+
+TIME_ORDER_MENU = """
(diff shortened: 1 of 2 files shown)


ReshefElisha:
CI fails on trigger '2' (checking 'top' category), I assume because the check doesn't expect the extra menu that was added?

michael-lazar:
Thanks, I'll try to look over this within the next couple of days. The tests will need to be updated, here's where the failure is https://github.com/michael-lazar/rtv/blob/master/tests/test_page.py#L82 and we will need to add a new test or two to cover the added functionality.

michael-lazar:
I figured out that there was an unrelated bug with sorting **/r/front** so I fixed it on this branch and re-enabled it. I also fixed the tests and added a couple of new ones. Thanks for the PR!

PR #314, opened 2016-10-10
Open a submission from `/` prompt

ReshefElisha:
Added the ability to open the comments page of a submission from the `/` prompt, given a correct URL. 

Changed behavior:
-  none, all valid input from previous versions should still be valid

Added behavior:
-  if a valid '/comments/' submission URL is given in the `/` prompt, the proper submission page will open

Possible improvements:
- `redd.it` URL parsing

PS: not sure what's up with line 278 (previously 275), a return symbol got automatically removed by my editor (vim-gnome)

### DIFF
--- a/rtv/subreddit_page.py
+++ b/rtv/subreddit_page.py
@@ -74,10 +74,10 @@ class SubredditPage(Page):
 
         name = self.term.prompt_input('Enter page: /')
         if name is not None:
-			if not '/comments/' in name:
-				self.refresh_content(order='ignore', name=name)
-			else:
-				self.open_submission(url='https://www.reddit.com/r/'+name)
+            if not '/comments/' in name:
+                self.refresh_content(order='ignore', name=name)
+            else:
+                self.open_submission(url='https://www.reddit.com/r/'+name)
 
     @SubredditController.register(Command('SUBREDDIT_FRONTPAGE'))
     def show_frontpage(self):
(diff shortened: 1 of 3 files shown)


rtv/subreddit_page.py
```diff
@@ -74,7 +74,10 @@ def prompt_subreddit(self):
 
         name = self.term.prompt_input('Enter page: /')
         if name is not None:
-            self.refresh_content(order='ignore', name=name)
+            if not '/comments/' in name:
+                self.refresh_content(order='ignore', name=name)
+            else:
+                self.open_submission(url='https://www.reddit.com/r/'+name)
```
michael-lazar: What's the intended format for typing in comment urls? The first time I tried this I went to https://www.reddit.com/r/pics/comments/571dw3 by typing
```
/r/pics/comments/571dw3
```
And it failed because it took me to
```
https://www.reddit.com/r/r/pics/comments/571dw3
```

Also, interestingly you don't need to put the subreddit name in the url so I could have also done https://www.reddit.com/comments/571dw3
```
/comments/571w3
```
That case also would also fail because ``name`` doesn't keep the slash at the start.

ReshefElisha: Hadn't thought of that. I'll add some better parsing/look to see if PRAW has something built in.

michael-lazar:
This PR has fallen out of date with the master branch. However, I really like the idea behind it so I made a new branch with a couple of changes based on my comments #340

Issue #290, opened 2016-08-26
[Feature Request] Ability to view subreddit side bar  


perplexedMurfy:

alexk307:
@perplexedMurfy @michael-lazar going to take a stab at this.

alexk307:
@michael-lazar three questions:

1. Is there a way to get the current size of the terminal?
2. Is there anyway to render markdown currently? If not, are there any libraries you would prefer me to use? All of the subreddit sidebars are in markdown and look terrible without proper rendering.
3. Which hotkey to use to pull up the sidebar? I was thinking `b`?

michael-lazar:
@alexk307 

good luck!

1. Yea that's easy to do with curses, see here for an example: https://github.com/michael-lazar/rtv/blob/master/rtv/page.py#L355. What are your thoughts on how you're going to show the info? Is it going to open a separate page, be a sidebar on the current page, a popup, or something else?
2. There are no markdown renderers for curses that I know of. It's actually a separate issue #274. I think the best that you can do is take fields that the api provides like n_subscribers and n_active_accounts and format them manually. 
3. ``b`` is already taken. I like ``i`` for "info" but that's taken as well. ``y``, ``t``, and ``g`` are open. It's looking like the inbox may never materialize so taking ``i`` back for this might be an option at some point in the future.

alexk307:
Sweet, yeah I can try to parse as many fields as possible. Currently I have it implemented as a notification that just overlays on the current screen. It might be more readable if it was in it's own pane like the help/options pane is. Ok I'll pick one of those letters!

alexk307:
This is what it looks like so far: ![image](https://cloud.githubusercontent.com/assets/3705729/20039389/14fe5ed2-a411-11e6-817c-3746094ae74d.png)

From what I can tell, the reddit API only returns a Markdown string as the sidebar/description. The only thing I think we could do is to parse the Markdown into some sort of better looking string.

michael-lazar:
Yikes! Yea that might be a roadblock unless you can figure out some way to parse that

alexk307:
My only thought is to use a markdown processor that converts markdown to html, then use beautifulsoup to parse the HTML into something readable.

I'll do some more researching after I get over this election...

alexk307:
@michael-lazar Parsed out the markdown which looks a little bit better. Not sure if there's any easy way to make it look any better on the terminal.
[...]

PR #324, opened 2016-11-13
#290 Ability to display subreddit sidebar

alexk307:


alexk307:
hmmm I added `Markdown` to `requirements.txt` but travis never installed it...Is there anything else I have to do to get it to install all requirements? @michael-lazar

michael-lazar:
Yea, you need to add it to setup.cfg and setup.py.

michael-lazar:
I'm closing this because IMO the proposed approach of stripping the sidebar text with no formatting at all produces an unreadable blob for most subreddits. The issue is still open if anybody want to come back to it.

PR #351, opened 2017-02-27
Small code improvement, using exist_ok=True

nix-py:
Using `os.makedirs(filepath, exist_ok=True)` instead 2 lines of if
block.


michael-lazar:
I would like to add this, but unfortunately that parameter was added in python 3.x and we need to support at least back to python 2.7. You can see this in the failing Travis CI build https://travis-ci.org/michael-lazar/rtv/jobs/205695887

Issue #346, opened 2017-02-25
Please add hotkey for copy url


averrin:
I was very surprised when `y` didnt copy url... =(

michael-lazar:
You mean copy it to the clipboard? Can you explain your use case further?

averrin:
Yes, to clipboard. Use case is simple - I want to share it. And before send it by messenger I need copy url.

noahfx:
Hi, 
I have implemented a solution for this https://github.com/noahfx/rtv/commit/7a2d248875918a69a1fb5bd1e39a4fae4bdc5ebf
@michael-lazar the reason why I have not made a PR yet is because it seems the library (pyperclip) fails sharing the clipboard outside the python application if there isn't some kind of clipboard manager installed, I think it is related with this https://github.com/asweigart/pyperclip/issues/61.

Do you have any suggestion? Any other library?

shaggytwodope:
A few thoughts towards @noahfx's solution. Assuming it's not changed, this would introduce some optional depends for packaging. And I personally think while this functionality is certainly welcomed. It would be ideal to be disabled by default. With packaging for AUR this is no problem. However many people install via PIP.

michael-lazar:
Thanks for working on this issue! I'm fine with adding a section in the README that says you need to install ``xclip`` or ``xsel`` separately to get it to work on linux. It would be similar to what ended up happening with ``urlview``. We would need to add a descriptive error message if you press the key but the appropriate program isn't installed.

That being said, it looks like outside of supporting the Windows clipboard, all that pyperclip is doing is piping commands to subprocesses. Do you think it would be better to avoid adding another python dependency and just call the commands directly?

noahfx:
Hi, 
I removed the pyperclip dependency and implemented my own copy :) 
Here is the PR of the feature https://github.com/michael-lazar/rtv/pull/364

michael-lazar:
Merged in #364

rigopoui:
This issue persists for wayland. My guess is that internally the program uses xclip, and not wl-copy

PR #364, opened 2017-05-04
Feature/copy url

noahfx:
Implementation of hotkeys to copy submission permalink and submission url, this resolves issue https://github.com/michael-lazar/rtv/issues/346


michael-lazar:
Thanks, I really appreciate what you did with the clipboard.py file! 

Comments:

- pyperclip still needs to be removed from setup.py
- unused ``sys`` import
- we need to add the new key bindings to the ``CONTROLS.rst`` document as well.
- Doesn't work for selfpost links (e.g. it copies the display name "self.explainlikeimfive" instead of the actual URL).  This should be an easy fix by using ``data['url_full']`` instead of ``data['url']``.
- You should use ``six.text_type()`` instead of ``str()`` to be compatible with python 2, otherwise ``.format()`` is mixing unicode with bytes which I'm pretty sure raises an exception. Actually, I think you can get away with sending ``.format(e)`` the raw error object and it will call ``__unicode__`` implicitly.
- I think it would be useful to flash a "Copied to clipboard" notification on success to give some user feedback that it worked, maybe with a short timeout of ~1 second.
- Needs some unit tests. I understand that it's difficult to setup the test environment for this project, so if necessary this is something that I can do myself before publishing the next release. For example I didn't test it, but theoretically if the subprocess raised an OSError it would crash the entire app.

noahfx:
I have changed all you asked and added a test
Let me know if the test is ok, by now it is passing in my local enviroment not sure if I wrote it in the right test file  `tests/test_submission.py`

michael-lazar:
I made a couple of minor changes and went ahead and merged your branch. Let me know if you have any questions.  I tried it out on my mac and found that ``os.name == 'mac'`` didn't actually work (it still returns ``posix``), so I switched to using ``sys.platform`` because it was already being used elsewhere in the codebase. The other big thing that I did was split your test into 2 separate test cases, one for OS X and one for Linux. I also altered the test so that it could still pass without 'xsel' or 'xcopy' installed, which is why it was failing on Travis. Thanks for the PR!

PR #367, opened 2017-05-28
Use imgur api to retrieve direct links to images.

woorst:
Entire imgur albums can be downloaded rather than the max 10 images from scraping the webpage.
There is a rate limit on the number of request. I'm not sure if there are enough users of `rtv` to run into any trouble.

### DIFF
--- a/rtv/mime_parsers.py
+++ b/rtv/mime_parsers.py
@@ -4,6 +4,7 @@ import mimetypes
 
 import requests
 from bs4 import BeautifulSoup
+import json
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


michael-lazar:
This is really interesting! I remember looking through Imgur's API a while ago but I didn't see the bit about non-oauth client id tokens. I couldn't find anything in the docs explicitly saying that the **client ID** is public and can be openly shared, but this part looks promising:

> We also limit each user (via their IP Address) for each application, this is to ensure that no single user is able to spam an application. This limit will simply stop the user from requesting more data for an hour.

It seems like the intent is for the **client ID** to be included in something like client-side javascript that gets executed by the end-users. So I think it's OK for rtv to use it in the same way.

I definitely think that the token needs to be added to the **rtv.cfg** though instead of being hardcoded in the source code. The default token can be shipped with rtv, but people should be able to swap in their personal API token just like they can do with reddit OAuth.

rtv/mime_parsers.py
```diff
@@ -104,64 +104,40 @@ def get_mimetype(url):
...
-            return " ".join(urls), 'image/x-imgur-album'
-        else:
-            return url, None
+        endpoint = 'https://api.imgur.com/3/{domain}/{page_hash}'
+        header = {'authorization': 'Client-ID {}'.format('d8842d573e8b9dd')}
+
+        pattern = re.compile(r'https?://(w+\.)?(m\.)?imgur\.com/((?P<domain>a|album|gallery)/)?(?P<hash>.+)$')
+        m = pattern.match(url)
+        page_hash = m.group('hash')
+        domain = 'album' if m.group('domain') in ['a', 'album'] else 'gallery'
```
woorst: There doesn't seem to be a way to reliably distinguish between gallery and image pages just from the url:

gallery: [https://imgur.com/NAVdRye](https://imgur.com/NAVdRye)
image: [https://imgur.com/0FiXBEw](https://imgur.com/0FiXBEw)

So I assume a given url is for a gallery page, which appear to be more common, and fallback to an image request upon failure.

rtv/mime_parsers.py
```diff
@@ -104,64 +104,40 @@ def get_mimetype(url):
...
-            img = div.find('img')
-            src = img.get('src') if img else None
-            if src:
-                urls.append('http:{0}'.format(src))
-
-        if urls:
-            return " ".join(urls), 'image/x-imgur-album'
-        else:
-            return url, None
+        endpoint = 'https://api.imgur.com/3/{domain}/{page_hash}'
```
woorst: I haven't had any trouble with my scheme did you? 
The following seem to be equivalent:
`https://api.imgur.com/3/gallery/{page_hash}`
`https://api.imgur.com/3/gallery/album/{{galleryHash}}`

---
`https://api.imgur.com/3/gallery/image/{{galleryImageHash}}`
This is if you want to request only a specific image within a gallery. Doesn't seem relevent to us.

michael-lazar:
Thank you for replying to my questions.

- I added the ``imgur_client_id`` to the configuration
- I brought back the imgur HTML scraper to act as a backup if the API fails or if our key is revoked. This is partially me being paranoid, but I don't trust the Imgur platform enough to rely 100% on their API being stable. 
- I added some extra tests

allthroughthenight:
Haven't updated rtv in a while, but looks like Imgur albums larger than 10 work with feh now. I'm guessing Imgur changed their API? This is to follow up to my issue [#399](https://github.com/michael-lazar/rtv/issues/399) which was a duplicate of this one.

PR #368, opened 2017-05-28
Create mime parser super class. Add streamable mime parser.

woorst:


michael-lazar:
Thanks! I'll need to test this out but it looks promising and I like the abstraction that you added.

Issue #369, opened 2017-05-28
Crash when trying to view saved comment.


woorst:
From the saved posts list I can view saved submission but rtv crash when I try to view a saved comment.
```
Traceback (most recent call last):
  File "/usr/bin/rtv", line 11, in <module>
    load_entry_point('rtv==1.13.0', 'console_scripts', 'rtv')()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 560, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2648, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2302, in load
    return self.resolve()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2308, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3.6/site-packages/rtv/__main__.py", line 169, in <module>
    sys.exit(main())
  File "/usr/lib/python3.6/site-packages/rtv/__main__.py", line 152, in main
    page.loop()
  File "/usr/lib/python3.6/site-packages/rtv/page.py", line 78, in loop
    self.controller.trigger(ch)
  File "/usr/lib/python3.6/site-packages/rtv/objects.py", line 591, in trigger
    return func(self.instance, *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/rtv/subreddit_page.py", line 126, in open_link
    self.open_submission()
  File "/usr/lib/python3.6/site-packages/rtv/subreddit_page.py", line 103, in open_submission
    url = data['permalink']
KeyError: 'permalink'
```

michael-lazar:
Thanks for creating the issue, it looks like an oversight that should be fairly easy to fix.

PR #373, opened 2017-06-02
Load comments

woorst:
Fix issue #369 
Also, can now see both comments and submissions on user pages.


michael-lazar:
Thanks this is great, I wonder why nobody thought of this earlier!

I added new cassettes for the failing tests. I also added a ``[Comment]`` tag at the beginning of comments so they're easier to distinguish from submissions.

Issue #363, opened 2017-04-19
Manage subscriptions


arvedarved:
I tried to subscribe to a subreddit from rtv and couldn't find it. 

Similar unsubscribing seems to be missing. 

Either it is undocumented or not implemented :).

As a workaround opening the reddit in  a browser (to subscribe it) would also be useful

michael-lazar:
You're correct that it's not currently implemented, it was requested a long time ago #193. I think it's worth adding but it hasn't been a high priority for me.

Issue #193, opened 2016-02-04
[Feature Request] Suscribe to subreddit from subreddit mode


luismichel:

michael-lazar:
I would like to avoid adding a new keybinding on the subreddit page, but when #392 gets merged you should be able to manage your subscriptions from inside of rtv.

PR #379, opened 2017-06-11
Add ability to manage subscriptions

woorst:
Addressing issues #363 #193.

This is not quite ready to merge. Right now you can:
- subscribe to a subreddit (press `/` while in the subscription page)
- unsubscribe to a subreddit (press `d` while in subscription page)
- create a multireddit (press `/` while in the multireddit page)
- delete a multireddit (press `d` while in the multireddit page)
- view contents of a multireddit (press `space` while in multireddit page)

Problems:
- I'm getting some sort of authentication error whenever I try to add/remove a subreddit from a multireddit
- listing the contents a multireddit is *very* slow. Much slower than actually opening multireddit
- ~~because you can't view an empty page in rtv, you can't navigate to a newly created multireddit and add subreddits to it~~
- idk what `modhash` is. I had to do a little hack to get that to work
- ~~do you think there should be a confirmation dialog before deleting an entire multireddit?~~

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -697,7 +697,7 @@ class SubscriptionContent(Content):
     @classmethod
     def from_multireddit(cls, reddit, loader, multi):
         items = iter(multi.subreddits)
-        content = cls('Multireddit: {}'.format(multi.path), items, loader)
+        content = cls('My Multireddit: {}'.format(multi.path), items, loader)
         content._multireddit = multi
         return content
 
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -694,6 +694,13 @@ class SubscriptionContent(Content):
 
         return cls(name, items, loader)
 
+    @classmethod
+    def from_multireddit(cls, reddit, loader, multi):
+        items = iter(multi.subreddits)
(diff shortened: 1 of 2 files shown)


michael-lazar:
Thanks, I've been a little busy lately but I'll checkout the branch and try it out when I get a chance

> - listing the contents a multireddit is very slow. Much slower than actually opening multireddit

You should check the rtv log and see which API requests are actually being made. Probably, the API call that returns the subreddits in a multireddit doesn't return all of the properties that the subscription page uses to draw the page. So PRAW silently makes a separate API request for each subreddit to fill in the missing properties.

> -because you can't view an empty page in rtv, you can't navigate to a newly created multireddit and add subreddits to it

Hmm yea I can't think of a good solution for this off the top of my head

> - idk what modhash is. I had to do a little hack to get that to work

https://www.reddit.com/dev/api/ says that it's for CSRF protection, but that it isn't required when using OAuth credentials so it sounds like it's deprecated. We might just want to set it to an empty string when praw is initialized and see if that breaks anything else.

> - do you think there should be a confirmation dialog before deleting an entire multireddit?

I think that sounds reasonable, anytime something is deleted there should be a confirmation dialog.

Issue #343, opened 2017-02-04
URxvt terminal flashes everytime you go up or down an entry


Hund:
Is it possible to somehow prevent this from happening in rtv? I'm actually using the flash-hint-thingy feature all the time in URxvt, so I don't want to disable it in URxvt. But it gets really annoying when you browse Reddit in rtv and it flashed every time you go up/down an entry. :)

**Edit:** I assume it's the visual bell feature that's causing it? I tried lxterminal and I didn't have the issue there.

michael-lazar:
It should only flash when you reach to first or last entry on the page and try to scroll out of bounds. If it's flashing every single time you move the cursor then something is for sure going wrong. Have you noticed any other strange behavior on URxvt? There might be an incompatibility somewhere between URxvt, your python curses build, your $TERM setting, or something else.

Hund:
I haven't noticed anything else with URxvt.

I'm running Arch and installed it via pip due to the issue with one of the dependencies in the AUR package. 

I'm not sure where to go from here, but let me know if you need anything or want me to do anything. I'll try to debug it myself the best I can as well.

dbaio:
Hi.
I noticed this as well on 1.14.1.
I've reinstalled rtv 1.13.0 and the flash's were gone.

dbaio:
I was using TERM=xterm-256color, replacing to TERM=rxvt-256color not solved the problem.

Commenting out this line stopped the flash's:

https://github.com/michael-lazar/rtv/commit/f3deea5abc06f7b22830a313f90929cbd8eac0b0#diff-abdf3d507d4d4d9968cb742523b25f8fR351

michael-lazar:
Thanks for the info @dbaio. I just installed rxvt and tested it out, here's what I observed right away: 

I did see the screen was flashing every time I scrolled, but it wasn't a visual bell. The difference is that with the actual visual bell command, the screen inverts colors for a second. But in this case it just looked like the screen was erasing and redrawing, and it was happening slow enough that it was visible to the eye. 

Does this sound like what you're seeing? If it is, it makes sense because the v1.14.1 update changed how the screen was getting redrawn and may have caused the slowdown. It's funny that this it only manifests in rxvt though (at least so far).

dbaio:
>  it just looked like the screen was erasing and redrawing, and it was happening slow enough that it was visible to the eye.

This makes sense.

There are a lot of issues about urxvt and redraw... I will do more tests tonight at home.

[...]

PR #383, opened 2017-06-21
Issue #343 Added method to clear screen

rreno:
School is over so I finally got around to this. The clear_screen() method in the Terminal class checks the $TERM variable in an attempt to avoid both the tmux and urxvt bugs. I was unable to test with tmux but it removes the flashing using urxvt.

### DIFF
--- a/rtv/terminal.py
+++ b/rtv/terminal.py
@@ -59,6 +59,7 @@ class Terminal(object):
         self.loader = LoadScreen(self)
         self._display = None
         self._mailcap_dict = mailcap.getcaps()
+        self._term = os.environ['TERM']
 
     @property
     def up_arrow(self):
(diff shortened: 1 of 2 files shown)


michael-lazar:
Thanks, I'll test this out when I get a chance

rreno:
I'm not really a fan of this upon looking at it again. Doing a check against an environment variable every screen refresh is bad.

I can add a class attribute that checks the variable when the class is initialized and just do the conditional check against that attribute in the method. Or maybe I'll think of something even cooler today.

rtv/terminal.py
```diff
@@ -769,3 +769,11 @@ def strip_textpad(text):
 
         out = '\n'.join(stack)
         return out
+    
+    # Resolves tmux touchwin() bug and urxvt clearok() flashing bug
+    def clear_screen(self):
+        if os.environ['TERM'] is not 'xterm-256color':
```
michael-lazar: You should be using ``!=`` for string comparisons

michael-lazar:
@codebam I'm not too concerned about it, string comparisons in python take negligible time and it's easier to follow along with the code the way it is.

@rreno I tried out your branch in rxvt/tmux and it looks to be working. I guess the only edge-case would be if somebody uses **rxvt** but sets their ``$TERM=xterm-256color``. At least now, they'll be able to fix it by switching to something more appropriate like ``$TERM=rxvt-unicode-256color``

Issue #393, opened 2017-07-15
[feature request] sort subreddit search by new, relevance, top, comments


sodiumjoe:
You can do this in the web ui. Currently, if you do a subreddit search (`f`), then sort by "new" for example, (`4`), rtv returns to the subreddit's new page.

michael-lazar:
Agreed, this would be nice to have.

michael-lazar:
The sorting options will now be applied to the current search, instead of discarding the query and refreshing the page.

Currently you can sort by (``hot``, ``new``, ``top``, ``rising``, ``controversial``). I opened a new issue to change the options to (``new``, ``relevance``, ``top``, ``comments``) for search results. https://github.com/michael-lazar/rtv/issues/401

PR #395, opened 2017-07-18
Search sorting

woorst:
#393

### DIFF
--- a/rtv/content.py
+++ b/rtv/content.py
@@ -435,10 +435,12 @@ class SubredditContent(Content):
     list for repeat access.
     """
 
-    def __init__(self, name, submissions, loader, order=None, max_title_rows=4):
+    def __init__(self, name, submissions, loader, order=None,
+                 max_title_rows=4, query=None):
 
         self.name = name
         self.order = order
+        self.query = query
         self.max_title_rows = max_title_rows
         self._loader = loader
         self._submissions = submissions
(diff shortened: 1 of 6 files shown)


michael-lazar:
I made some changes and since I can't commit to your branch, I made a [new PR](https://github.com/michael-lazar/rtv/pull/400). Please check it out and let me know what you think. The changes that I made are as follows:

- Added the search result to the page title (https://github.com/michael-lazar/rtv/issues/396)
- Was there a reason that you only ordered search results for **new** and **controversial**? I tried some queries and was able set the order to **hot**/**top**/**rising** without errors. 
- Got rid of the requirement for a  ``with_search`` argument by adding this check to ``SubredditPage.refresh_content()``. This results in a smaller change to the codebase, and I can't think of any examples where it fails compared to ``with_search``.

    ```python
    # Preserve the query if staying on the current page
    if name is None:
        query = self.content.query
    else:
        query = None
     ````

woorst:
Your changes look good.

I had only ordered search results by **top**, **new** and **controversial** because these are the only options available when accessing the search function via a web browser. I wasn't sure if ordering by **hot** or **rising** was returning sensible results. But I'm fine with how you've changed it.

michael-lazar:
Thanks for looking it over. I didn't realize that the sorting options were different, I opened a separate issue for that.

Issue #372, opened 2017-06-02
[Feature Request] Improve comment navigation


woorst:
A shortcut key to scroll to a comment's parent.

michael-lazar:
Any suggestions? I think it would need to include a keyboard modifier, like <kbd>shift</kbd>+<kbd>j</kbd> / <kbd>shift</kbd>+<kbd>k</kbd>?

PR #397, opened 2017-07-19
jump directly to parent/sibling comments in submission page

woorst:
#372


michael-lazar:
Thanks! Code looked great, the only thing that I added was terminal flashes() for invalid <kbd>J</kbd>/<kbd>K</kbd> presses.

DailytheNoob:
Question: how do I modify this to use the keys that I set in the config instead of J/K?

Issue #389, opened 2017-07-09
Missing COPY_PERMALINK in [bindings]


frdrcv:
When starting rtv, I see the "logging..." box and I get the following error:
`"Invalid configuration! `COPY_PERMALINK` key is undefined"`

I'm using rtv 1.16.0 on Arch Linux, installed via the AUR.

michael-lazar:
You probably have custom key bindings setup in your ``~/.config/rtv.cfg`` file. There was a new command added in a recent version, but its not defined in your configuration. To fix this:

1. Delete the [bindings] section in your ``~/.config/rtv.cfg``, or
2. Add the COPY_PERMALINK command to your [bindings] section

frdrcv:
Thanks, the second solution fixed my problem.

michael-lazar:
Glad it worked! I'm going to leave this issue open to remind myself that custom bindings need to be handled better by the code. It shouldn't break everyone configuration every time a new binding is added.

PR #398, opened 2017-07-19
set_bindings can now update or replace existing bindings

woorst:
#389

### DIFF
--- a/rtv/objects.py
+++ b/rtv/objects.py
@@ -642,16 +642,16 @@ class KeyMap(object):
         self._keymap = None
         self.set_bindings(bindings)
 
-    def set_bindings(self, bindings, mode='update'):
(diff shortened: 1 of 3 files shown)


michael-lazar:
I don't think it's necessary to provide both "update" and "replace" modes, since it appears that the code will only ever use "update". What about simplifying the code to always update the keymap?

```python
class KeyMap(object):
    """
    Mapping between commands and the keys that they represent.
    """

    def __init__(self, bindings):
        self._keymap = {}
        self.set_bindings(bindings)

    def set_bindings(self, bindings):
        for command, keys in bindings.items():
            if not isinstance(command, Command):
                command = Command(command)
            self.keymap[command] = keys
```

michael-lazar:
Thanks!

Issue #415, opened 2017-08-18
Add [S] to Submitter's username in comments


cmccandless:
# Issue
Submitter's username on comments posted by the same is indistinguishable from other users when user --monochrome.

## Example:
_In below screenshots, user cheese9k is the Submitter._
### Without --monochrome:
![color](https://user-images.githubusercontent.com/7275238/29466049-12e7b9ec-8401-11e7-8f1c-0a94c254afcf.PNG)

### With --monochrome:
![monochrome](https://user-images.githubusercontent.com/7275238/29466053-162cd092-8401-11e7-9c83-adddb693e87f.PNG)

# Proposed Solution
Add "[S]" to end of Submitter's username on comments posted by Submitter. This is how Reddit handles this case, as well as changing the color of the username (see below).

### From Browser:
![browser](https://user-images.githubusercontent.com/7275238/29466055-1820a810-8401-11e7-8634-381f8748374a.PNG)

michael-lazar:
+1 I think this is a good suggestion, especially if it's something that Reddit is already doing on their site. 

This should be a very simple PR if somebody wants to take it on. Here's where the author text is being drawn: https://github.com/michael-lazar/rtv/blob/master/rtv/submission_page.py#L265

cmccandless:
I've got it covered... pull request forthcoming.

PR #416, opened 2017-08-18
RE #415: Add '[S]' to submitter username on comments

cmccandless:
Per #415, add '[S]' to the end of the submitter's username on their own comments.


michael-lazar:
Thanks!

PR #450, opened 2017-10-07
handle single image imgur galleries

woorst:
Imgur galleries containing only a single animated image were not being parsed properly causing a crash.
example: https://imgur.com/gallery/vLHI7


michael-lazar:
Thanks!

PR #456, opened 2017-10-15
Added additional installation way

mehandes:
Added method using Nix package manager


michael-lazar:
Thanks :) I've never used NixOS, so I'm glad that you were able to test it.

PR #457, opened 2017-10-16
Youtube videos are now directly handled by mpv in mailcap

anhtuann:
I simplified the mailcap entry for youtube videos using mpv. No need to pipe the stream from youtube-dl into mpv anymore since this latter can directly handle youtube urls. Bonus side effect : it's now possible to navigate backward and forward in the video.


michael-lazar:
Thanks!

PR #458, opened 2017-10-18
Changed gif handler to mpv and removed redundant webm entry in mailcap

anhtuann:
Gifs were previously handled by feh which can only display the first frame.


michael-lazar:
Thanks!

Issue #474, opened 2017-11-06
Word wrapping in $PAGER for long text


jupart:
Currently, rtv pipes text for a submission, comment, etc. directly into $PAGER (like `less`). `less` breaks lines based on length, ignoring/not caring if it breaks a word. I'd like to add functionality to break lines without breaking words, I'm thinking about piping the text into `fold` and piping that into $PAGER. Comments, opinions, objections?

michael-lazar:
I like the idea. RTV already has the capability to wrap text based on words, in fact it probably does a better job that ``fold`` because it's aware of unicode-wide characters. It might be easier to do this in python.

PR #479, opened 2017-11-13
Wrap text in submission page PAGER output

jupart:
Issue #474

### DIFF
--- a/rtv/terminal.py
+++ b/rtv/terminal.py
@@ -552,7 +552,7 @@ class Terminal(object):
             with self.suspend():
                 webbrowser.open_new_tab(url)
 
-    def open_pager(self, data, wrap=False):
+    def open_pager(self, data, wrap=None):
         """
         View a long block of text using the system's default pager.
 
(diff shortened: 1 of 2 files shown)


rtv/terminal.py
```diff
@@ -552,7 +552,7 @@ def open_url_silent(url):
             with self.suspend():
                 webbrowser.open_new_tab(url)
 
-    def open_pager(self, data):
+    def open_pager(self, data, wrap=False):
```
michael-lazar: Just for convention's sake I think the default should be ``None`` instead of ``False``

rtv/terminal.py
```diff
@@ -561,6 +561,14 @@ def open_pager(self, data):
 
         pager = os.getenv('PAGER') or 'less'
         command = shlex.split(pager)
+
+        if wrap:
+            data_lines = content.Content.wrap_text(data, wrap)
+            data = ''
+            for line in data_lines:
```
michael-lazar: You could simplify this to ``data = '\n'.join(data_lines)``

Issue #491, opened 2017-12-27
Put history.log and refresh-token in $XDG_DATA_HOME


calebperkins:
The `~/.config/rtv/refresh-token` and ` ~/.config/rtv/history.log` files could be written to `$XDG_DATA_HOME` (or possibly `XDG_CACHE_HOME`) instead of `$XDG_CONFIG_HOME`. It's a common practice to version control the config directory, and the XDG specification makes this distinction between config and data files, which change more regularly.

michael-lazar:
This has been implemented

PR #492, opened 2017-12-27
Use XDG_DATA_HOME for history and refresh token.

calebperkins:
Implementation of #491. This moves the history and refresh-token files into XDG_DATA_HOME, so that the program's config and transient program data are separated.


michael-lazar:
Thanks for the PR, I'm on board with this change. Do you think we should stick the history file into XDG_CACHE_HOME? It sounds like a more appropriate directory, as long as the file is being moved either way.

calebperkins:
I did some more Googling for what belongs in data versus cache, I think it's fine to remain in XDG_DATA_HOME.

Based on [GNOME's recommendation](https://wiki.gnome.org/action/show/Initiatives/GnomeGoals/XDGConfigFolders):

> * If you remove the .cache folder, the user MUST not notice it except maybe for performance (might be huge performance drop)
> * .local should be treated as any other user data

And [Debian's recommendation](https://wiki.debian.org/XDGBaseDirectorySpecification):

> The above example information is not essential data. However it should still persist on reboots of the system unlike cache data that a user might consider putting in a TMPFS.

So it seems to be that the cache directory is for anything the program can regenerate without user intervention, and the data directory is for anything else.

michael-lazar:
Good info, thanks again!

Issue #326, opened 2016-11-30
[Feature Request] Ability to hide links


data-flux:
Thanks for this amazing program!

I have only one problem with it, I like to hide links, when I've seen them.

As far as I can tell, this is not currently possible.

If at all possible, could you add this feature?

data-flux:
Clarification, this is what I hope to do:

https://www.reddit.com/dev/api/#POST_api_hide

michael-lazar:
It looks like there's a function exposed through PRAW here https://github.com/praw-dev/praw/blob/praw3/praw/__init__.py#L2605 ,so I don't think that this would be very difficult at all. Are you thinking of having a config option to auto-hide links that were visited, or having a key that you can press to hide the selected link?

data-flux:
Keypress for sure.

armandg:
This featur would be great! Any chance of setting this so you could chose auto-hide or hide by keypress? Perhaps keypress also should hide/show the posts you already read?

nemanjan00:
Any news on this?

michael-lazar:
No, there hasn't been any work done on this yet. Since there have been several people commenting / voting on this issue I added a high priority tag.

nemanjan00:
Ok, guys, I am working on PR. 

It would be great if someone could also test it. :)

Thank you!

armandg:
@nemanjan00 If I get some starting help with getting your PR down to my machine and running I will be more than happy to test =)

nemanjan00:
```bash
git clone git@github.com:nemanjan00/rtv.git
cd rtv
git checkout hide
python -m rtv
```

Button for hiding is x. 

If you have custom config, add: 

```
HIDE = x
```

above 

```
UPVOTE = a
```

armandg:
Both standard shortcut and changing in settings works great.

A few things that also should be fixed:

- The "hide" button should toggle hidden or not, just like for saving posts or comments (press `w` multiple times on a post, and you'll see what I mean).
- Should this function perhaps also be described in the bottom status bar with the others?

nemanjan00:
Ok, I will see what I can do.

armandg:
@nemanjan00 Also, trying to hide a comment crashes rtv and gives following output:

```Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
[...]

PR #498, opened 2018-01-11
Hide command #326

nemanjan00:
I have started implementing feature from #326. 

So far I have key binding that sends hide request but that is not reflected on GUI until refresh.


nemanjan00:
Seams like i broke tests for saved and have no idea now to fix them.

michael-lazar:
Thanks for the PR, I'll do a proper review within the next couple of days.

In the meantime, there some instructions on how to setup and run the tests here:

https://github.com/michael-lazar/rtv/blob/master/CONTRIBUTING.rst

nemanjan00:
@michael-lazar, setting up test enviroment is not as big of a problem as understanding tests sinch I do not know python...

nemanjan00:
So, can I get some help with this? :)

michael-lazar:
- https://travis-ci.org/michael-lazar/rtv/jobs/327698876#L1193

  This is failing because when you defined a new ``hidden`` param in the default theme, you also 
  needed to add it to the other theme files that are packaged with rtv (molokai, papercolor, solarized-dark, solarized-light).

- https://travis-ci.org/michael-lazar/rtv/jobs/327698876#L951
  https://travis-ci.org/michael-lazar/rtv/jobs/327698876#L1065
  https://travis-ci.org/michael-lazar/rtv/jobs/327698876#L1065

  These are all caused by the same underlying problem. Try running ``rtv -s /u/spez`` and you should see the error. The problem is that the subreddit page will sometimes display ``SavedComment`` objects, for example on the ``/u/saved`` or ``/u/me`` pages. https://github.com/michael-lazar/rtv/blob/master/rtv/content.py#L163 This type of object didn't ever define a ``data['hidden']`` param because it wasn't ever necessary before. You can fix it by adding ``data['hidden'] = False`` to the section of the code that I linked to.

- What do you think about using the <kbd>space</kbd> key to hide/unhide submissions? It's already being used to hide/expand comments, so I think it would make sense to overload the same keybind instead of adding a new one.

- I guess in order to unhide submissions after the page refresh, we would need to add a ``/u/hidden`` endpoint that returns all of your hidden submissions, similar to how ``/u/saved`` works with saved submissions. That's kind of a separate feature though so it doesn't need to be a part of this PR.

- Overall I like the approach of adding a ``[hidden]`` flair. Originally I was thinking that hiding a submission would delete it from the page altogether and refresh the screen. But this way it gives feedback, and allows the user to revert if they made a mistake.

nemanjan00:
Hi, 
Motherboard in laptop I used for working on this died and I am not going to continue working on it Until I replace it. 

So, you can expect progress in 5-6 days... 

Sorry for delay.

michael-lazar:
No problem, not in a rush

nemanjan00:
So, tests have passed. What is left is documentation and making decision if you want to use ``space`` or ``x``

nemanjan00:
Can we expect this to be merged? :)

michael-lazar:
Fixed the docs and some other minor things, added a unit test, and merged the branch. Thanks for being patient!

nemanjan00:
Great! Thanks! :)

PR #515, opened 2018-02-21
Fix theme name in a sample config file

klivan:
Fix theme name in a sample config file


michael-lazar:
Thanks! I always get those two confused...

PR #570, opened 2018-07-05
pager: Add RTV_PAGER to support a reddit-specific setting

joeythesaint:
Reddit submissions and comments are styled with Markdown.  If you have a
Markdown-aware pager/viewer on your system it is convenient to be able to
use it rather than the default 'less'/'more'/etc. one used for everything
else on the system.  Support a high-priority environment variable that
will be consulted prior to checking $PAGER and/or defaulting to 'less'.

Signed-off-by: Joe MacDonald <joe_macdonald@mentor.com>


michael-lazar:
yes xfailed means expected failure, nothing to worry about

PR #572, opened 2018-07-16
Add 'gilded' as a supported sort

edridgedsouza:
I added 'gilded' to the top-level sort in the banner, as well as in the accepted sorts in the command prompt


edridgedsouza:
I believe the source of these failed tests are the ones explicitly related to the hard-coded menu bar. In my own repo, I attempted to fix these in a new commit, though I can't verify whether the tests pass or not. I can vouch that the new 'gilded' feature works smoothly on my machine, though, so fixing the tests would likely make the new feature usable.

PR #583, opened 2018-08-06
Refactor __main__.py for readability

codesoap:
I know you explicitly asked people not to make style-only pull requests. I know it's a big commit.

But I think it improves readability considerably and I didn't change anything in the logic.

### DIFF
--- a/rtv/__main__.py
+++ b/rtv/__main__.py
@@ -60,18 +60,68 @@ _logger = logging.getLogger(__name__)
 def main():
     """Main entry point"""
 
-    # Squelch SSL warnings
+    config, user_agent = initialize()
+    if not config:
+        return
+
+    try:
+        with curses_session() as stdscr:
+            term = Terminal(stdscr, config)
+            run(config, user_agent, term)
+    except ConfigError as e:
+        _logger.exception(e)
+        print(e)
+    except Exception as e:
+        _logger.exception(e)
+        raise
+    except KeyboardInterrupt:
+        pass
+    finally:
+        # Try to save the browsing history
+        config.save_history()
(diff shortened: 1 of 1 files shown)


codesoap:
Darn, I had run the tests locally, but didn't see the .pylintrc *blushes*
I couldn't quickly come up with a fix for the problematic lines. I keep trying and hopefully be able to add another commit in the next days.

codesoap:
OK, looks like the PR is ready to be reviewed.

michael-lazar:
I'm sorry, but I disagree that this improves readability. Instead of reading the file from top-to-bottom, you now need to jump back and forth between a bunch of nested functions. The functions themselves don't provide any value over the block comments that they're replacing. Everything is already contained to a single file and there's not code duplication. Sometimes simple is better than complex.

I admit that this is a personal preference and people have different coding styles. There's nothing objectively *wrong* with this PR, the code looks good and it passes the tests. But it's not the direction that I want to go in.

codesoap:
Alright, thank you for explaining your standpoint. I can accept that you have a different taste in code.

I don't intend to change your mind about the PR, but in case you are interested in what benefits I see:

1. Without needing to understand all implementation details you can quickly get a rough idea about what's happening by reading the top-level functions.
2. If you are new to the code you can find the part of the code you're interested in quickly, by following function names.
3. It's easier to reason about (side)effects of code-changes, because you can see which variables are passed from and to functions.

My taste in code is strongly influenced by the book "[Clean Code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882)".

Issue #597, opened 2018-08-18
Improve the startup speed by not loading a subreddit page when rtv is launched with a submission link as parameter


codesoap:
I would like my URL resolver to open reddit links with rtv. Currently this is not really an enjoyable experience. I want to suggest these changes to rtv to improve the behavior in this scenario:

1. Improve the startup speed by not automatically authorizing the user. Instead the user could be authorized, if he/she tries to perform an action that requires authorization.
2. When I to leave the submission page I get to the subreddit view of rtv. When rtv was opened with a submission link as parameter I would prefer it to quit instead.
3. Improve the startup speed by not loading a subreddit page when rtv is launched with a submission link as parameter.

Remark: I have already implemented most of the suggested changes for myself and can make pull requests if these changes sound acceptable to you, @michael-lazar .

michael-lazar:
1. Beyond authenticated actions, the submission stream itself depends on if the user is authenticated. There's the customized front page, and there are also more subtle things like hidden submissions and showing nsfw content.  I would be open to adding a command line option to skip login on startup, but I don't want to make this the default behavior.

2. Why not just quit then with ``ctrl-c`` or ``q``? Changing the behavior seems like it would confuse more people than not.

3. I'm 100% for this, I agree that loading the subreddit page first is unnecessary. The only reason it was done that way is because it was easier.

codesoap:
1. OK, I was not aware of the subtle changes when logging in. I will open a PR for the introduction of a config parameter.

2. + 3. I was hoping you would be cool with 2., because implementing those two together makes things much easier :P. I understand your point, though. When I find some time I'll try to make a PR for 3. .

michael-lazar:
Number 3 was implemented in PR #658. Opening a submission link no longer requires loading the submission at startup.

PR #599, opened 2018-08-27
Issue #597: Autologin configuration

codesoap:
This PR introduces the `--no-autologin` cli parameter and the correlating `autologin` config option. A user could be willing to skip automatic login, to speed up the startup of rtv.


michael-lazar:
I tested this out and it appears to be working as intended, I like the new flag name. I made a change so that the ``"Welcome {user}!"`` message will still be displayed when they login when using the ``--no-autologin`` flag. I also added some test coverage.

Issue #615, opened 2018-10-01
Open links dialog: previous page option


woorst:
I like the new open links dialog. Can an addtional option be added so I don't have to close and reopen the prompt if I over shoot the page with the link I want?

Example: 
```
[9] next page...
[0] ...previous page
```

or, so that more links can be shown per page: 

```
[j] next page...
[k] ...previous page
```

michael-lazar:
Looks good to me 👍

PR #616, opened 2018-10-01
Previous page link in open links prompt.

woorst:
#615 

and up to 10 links visible on every page


michael-lazar:
Thanks for the PR, I think this is a solid improvement!

PR #626, opened 2018-11-03
open links prompt to select url for copying

woorst:


woorst:
@michael-lazar

michael-lazar:
Thanks woorst. I made the following changes and merged

- Added tests
- Changed the behavior to not flash the screen if the user is presented with the prompt but doesn't select a link.
- Made the code more tolerant of missing content fields. E.g. the same method was using both``data.get('permalink')``  and ``data['permalink']`` interchangeably.

PR #629, opened 2018-11-13
Add max column width for pager

micronn:
Some times I find it easier to read the text when it's wrapped to 70 columns. So I've added an option to set the maximum width (disabled by default) for the pager.


pltrz:
+1 for this. It's difficult to read text in the pager with really long lines.

michael-lazar:
@micronn Thanks for the PR, I think this is a reasonable addition . Sorry for the delay in getting it merged!

micronn:
No problem. Thanks for the merge! :)

PR #636, opened 2019-01-20
Mailcap: support URLs and files

no2chem:
The default mailcap for rtv is nice with some presets for viewing images in iTerm2, but it expects that paths are URLs. Other apps that use mailcap, such as mutt/neomutt, provide paths on the filesystem instead of URLs.

This PR modifies the mailcap template to check if the provided path starts with http, and uses curl to fetch the image only if the path starts with http. Otherwise, it directly pushes it to iterm2's imgcat. This makes it possible for mutt and rtv to share the same mailcap.


michael-lazar:
Thanks! I'm no expert in mailcap, and it's nice to have input from people who also have experience using it with other applications.

Issue #566, opened 2018-06-23
README should not advice pacaur but yaourt


pyluyten:
For Arch users : the README says to run pacaur. But according to this arch wiki page, this is problematic : https://wiki.archlinux.org/index.php/AUR_helpers

So README could instead say to run `yaourt -S rtv` or just point to repology.

pyluyten:
ouch sorry. So, i would just point to the AUR package without precising any instruction (after all the AUR wiki page does not tell to use any helper), but you are more qualified to decide.

michael-lazar:
I'll keep it as ``pacaur`` for now, and leave the issue open so others can weigh in.

jathanasiou:
Both `pacaur` and `yaourt` have issues as of now. The completely secure options are (in the order listed in arch wiki):

- aurman
- aurutils
- pakku
- pikaur
- trizen
- yay
- bauerbill


Out of these `aurman`, `trizen` and `bauerbill` can also be used as full pacman replacements while the former also fetches PGP keys automatically based on the table there. Personally, I've been using `pakku` as it pulls AUR comments as well, but its build syntax can be a bit cumbersome compared to yaourt/pacaur.  

Manjaro has also developed the graphical [pamac](https://gitlab.manjaro.org/applications/pamac).

mrcpj1998:
Adding on to this. Don't get dragged down in managers and politics, just point to rtv aur package

pabloariasal:
It is the norm to just provide a link to the package in the AUR, every single Arch user will know how to proceed. As @mrcpj1998 suggests, AUR installation instructions are pointless specially if they include unmaintained tools.

RakNoel:
Adding this: [https://www.reddit.com/r/archlinux/comments/7k5suz/pacaur_now_unmaintained/](url)

It might be time to change that, as it might be misleading for new arch-users. 

Whatever you do recommend is probably not all that important. Id say either do a) "yay rtv-git" becuase "yay" sounds fun and worked for me, b) simply give a link to the arch-wiki about AUR-helpers, or c) so like @meskarune and just make them build it. :package: 

Id go with yay :1st_place_medal:

Almamu:
I'd do the same as @pabloariasal and @mrcpj1998 said, give a link to the aur package. I would add a link to the aur documentation on archlinux just in case, but nothing more.

michael-lazar:
[...]

PR #639, opened 2019-01-29
Update Arch Linux installation instructions

Crestwave:
The Arch package has been moved to the [official community repository](https://www.archlinux.org/packages/community/any/rtv/).


michael-lazar:
Awesome, thanks for pointing this out!

Solves discussion in #566

PR #642, opened 2019-02-02
Improve comment formatting for markdown code block

woorst:
Lines beginning with four space are treated as code on reddit.
`rtv` strips leading whitespace before sending posts to reddit thus preventing code blocks from being recoginzed.
This fix should strip only leading empty lines rather than all leading whitespace.


michael-lazar:
Thanks! Added some test cases and merged.

Issue #647, opened 2019-02-06
Display Logout instead of Login in the FOOTER if authenticated user.


gcarrarom:
Whenever using RTV, the footer displays login even though I'm already logged in.
Suggestion: Add a second FOOTER definition in the docs.py and add the logic in the init for the page.

michael-lazar:
👍 Sounds like a minor and reasonable addition if somebody wants to open a PR

aiden-sobey:
I'm taking a shot at this one.

PR #648, opened 2019-02-06
Add Logout text in footer when user is logged in

gcarrarom:
Fix for issue #647


gcarrarom:
Closed as there's a bug on logout

PR #654, opened 2019-02-26
Footer displays 'login' or 'logout' when suitable.

aiden-sobey:
Fixes issue #647 

The correct term (login/logout) is now displayed in the footer, respective of if a user is currently logged in or not.

Note: I'm not a big fan of self.FOOTER being capitalised like a constant because.. well, it's not. But changing this broke functionality in the parent class so I decided to keep the diff small. If desired though I can go through and fix this up, the fix will just get a fair bit more complicated.

### DIFF
--- a/rtv/subreddit_page.py
+++ b/rtv/subreddit_page.py
@@ -33,12 +33,14 @@ class SubredditPage(Page):
         self.content = SubredditContent.from_name(reddit, name, term.loader)
         self.nav = Navigator(self.content.get)
         self.toggled_subreddit = None
+        self.FOOTER = self.reload_footer()
 
     def refresh_content(self, order=None, name=None):
         """
         Re-download all submissions and reset the page index
         """
 
+        self.FOOTER = self.reload_footer()
         order = order or self.content.order
 
         # Preserve the query if staying on the current page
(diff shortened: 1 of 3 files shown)


rtv/subreddit_page.py
```diff
@@ -60,6 +62,12 @@ def refresh_content(self, order=None, name=None):
         if not self.term.loader.exception:
             self.nav = Navigator(self.content.get)
 
+    def reload_footer(self):
```
michael-lazar: I think this would be a great place to use a property, e.g.

```python
@property
def footer(self):
    if self.reddit.user is None:
        return docs.FOOTER_SUBREDDIT
    else:
        return docs.FOOTER_SUBREDDIT_LOGGED_IN
```

Then all you would need to do is change the other pages to lowercase, e.g. for the subscription page

```python
footer = docs.FOOTER_SUBSCRIPTION
```

and update the base class to use ``self.footer`` instead of ``self.FOOTER``. I agree that uppercase no longer makes sense, and it was probably a questionable decision when it was added in the first place. If you want to change ``self.HEADER`` to lowercase at the same time I'm on board with it.

PR #670, opened 2019-03-31
Toggleable footer via config

marcusmoller:
I've made a few simple commits to allow for a toggleable footer keybinds help bar via the configuration file.


marcusmoller:
That's rather odd, I'm unable to reproduce that bug on new installation. What I did to test:

- clone my repository, install using `python setup.py develop` and checkout the footer_hide branch
- copy config to home dir using `rtv --copy-config`
- run rtv in terminal

Issue #669, opened 2019-03-30
Add option to open browser links in new window


pabloariasal:
I use a tiling window manager where I have a dedicated workspace only for rtv and usually have other browser instances opened in different workspaces. 

rtv always opens a new tab in an existing browser instance instead of opening a new browser window. Hence, I'm forced to change workspaces and look for the browser instance where the page was opened.

Ideally one could set a configuration parameter to alter this behavior (i.e. always open a new window instead of a new tab). This issue is related to #506, but less complex, as one may continue to use `webbrowser` module with a different call.

I would like to issue a pull request for this, if my proposed solution seems reasonable.

michael-lazar:
I think this is a reasonable solution. I don't think that it's significant enough to add a new command line flag, but a ``rtv.cfg`` setting makes sense.  Go for it!

PR #674, opened 2019-04-14
Add option to force opening links in new browser window

pabloariasal:
#669 

Added an option to the config file, _force_new_browser_window_, which when set opens links 
in a a new window instead of a new tab in the running browser. 

[Note] Option works with old config files. If the config file doesn't have this option defined the old behavior (open in new tab) will be performed.

### DIFF
--- a/rtv/config.py
+++ b/rtv/config.py
@@ -261,6 +261,7 @@ class Config(object):
         if config.has_section('rtv'):
             rtv = dict(config.items('rtv'))
 
+        # convert non-string params to their typed representation
         params = {
             'ascii': partial(config.getboolean, 'rtv'),
             'monochrome': partial(config.getboolean, 'rtv'),
@@ -274,7 +275,8 @@ class Config(object):
             'max_comment_cols': partial(config.getint, 'rtv'),
             'max_pager_cols': partial(config.getint, 'rtv'),
             'hide_username': partial(config.getboolean, 'rtv'),
-            'flash': partial(config.getboolean, 'rtv')
+            'flash': partial(config.getboolean, 'rtv'),
+            'force_new_browser_window': partial(config.getboolean, 'rtv')
         }
 
         for key, func in params.items():
--- a/rtv/templates/rtv.cfg
+++ b/rtv/templates/rtv.cfg
@@ -57,6 +57,9 @@ hide_username = False
 ; been installed into either the custom of default theme paths.
 ;theme = molokai
 
+; Open a new browser window instead of a new tab in existing instance
+force_new_browser_window = False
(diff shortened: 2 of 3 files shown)


rtv/terminal.py
```diff
@@ -599,7 +599,10 @@ def open_url_silent(url):
                     try:
                         os.dup2(null, 1)
                         os.dup2(null, 2)
-                        webbrowser.open_new_tab(url)
+                        if self.config['force_new_browser_window']:
```
michael-lazar: There's another place that ``webbrowser.open_new_tab()`` is called at the bottom of this method. That will also need to be updated for consistency.

rtv/config.py
```diff
@@ -261,6 +261,7 @@ def _parse_rtv_file(config):
         if config.has_section('rtv'):
             rtv = dict(config.items('rtv'))
 
+        # convert non-string params to their typed representation
```
michael-lazar: nice comment!

rtv/templates/rtv.cfg
```diff
@@ -82,6 +82,9 @@ oauth_scope = edit,history,identity,mysubreddits,privatemessages,read,report,sav
 ; See https://imgur.com/account/settings/apps to generate your own key.
 imgur_client_id = 93396265f59dec9
 
+; Open a new browser window instead of a new tab in existing instance
+force_new_browser_window = False
```
michael-lazar: This should be moved up, out of the ``OAuth Settings`` section and below ``;theme = molokai``.

michael-lazar:
Thanks for the PR. I made a few requests for minor changes.

rtv/terminal.py
pabloariasal: Oh ok I actually I didn't add it there on purpose because I thought that it doesn't make sense to do this for a console browser. I will add it.

rtv/config.py
pabloariasal: :)

rtv/templates/rtv.cfg
pabloariasal: Oops yeah missed that in fact!

pabloariasal:
Thank you for the review, @michael-lazar! I addressed the requested changes. I created separate commits because I'm not sure if one can do a force push here.

rtv/terminal.py
michael-lazar: No worries. There are actually edge cases where that code path gets executed with GUI browsers too (like if the user's ``$DISPLAY`` is set wrong).

michael-lazar:
Merged, thank you!

PR #675, opened 2019-04-17
Make sure feh scales the image to fit the window

mattydebie:
This is a real small PR for the mailcap template, but it's something I was missing when viewing images.
```bash
-., --scale-down                                                                                                                                                                                            
  Scale images to fit window geometry (defaults to screen size when no geometry was specified).
  Note that the window geometry is not updated when changing images at the moment.
  This option is recommended for tiling window managers.  This option is ignored when in 
  fullscreen and thumbnail list mode. 
  In tiling environments, this also causes the image to be centered in the window.
```


michael-lazar:
thanks!

PR #684, opened 2019-04-28
Add ability to move selected item on top (fixes #683)

### LINKED ISSUE #683: Improve navigation by adding the ability to put the currently selected comment on top of the page
**Is your feature request related to a problem? Please describe.**
When you move down a page, you reach the point where you don't see the upcoming items on the screen. It would be nice if we could have a new navigation feature to put the currently selected item on top of the visible window.

**Describe the solution you'd like**
A new key press event, that puts the currently selected item on top of the visible window.

**Describe alternatives you've considered**
Moving a whole page down (by pressing `n`) is somewhat similar to this request, but it currently has other problems, described in #682.

goggle:
As described in #683 this PR adds a new navigation feature to RTV, that allows to move the currently selected item on top of the visible window.


michael-lazar:
Thanks for the PR (and test!). I want to sit on it for a little while to ponder about the key binding, but overall your code additions look pretty clean and straightforward.

goggle:
Yes, I wasn't sure about the key binding either. It would be nice if it is something on the right side of the keyboard, since the other navigation keys are also on the right side.
There might also be a better name than `focus_up`.

jparkrr:
This is a dope feature. Don't know about the keybinding, but would love to see it merged.

goggle:
Yeah, on a Swiss QWERTZ keyboard to `-` key is really convenient to use for this, but probably not on an American QWERTY keyboard. I think the corresponding key on an American keyboard would be `/`. But this key is already reserved for the prompt.

Maybe the zero key `0` would be a good alternative?

